PR #12, opened 2022-06-17
refactoring

AnthonyBuisset:
- :recycle: revamp code to make it more modular
- :recycle: create single processor struct

modular architecture to respect SOLID principles
binary crates are plugging modules all together
next step would be to add tests


abdelhamidbakhta (approved):
lgtm

PR #22, opened 2022-06-22
:recycle: use iterator pattern in Fetcher API

### LINKED ISSUE #19: use iterator trait for in fetcher api
In order to interact easily with multiple results from `Fetcher` API, we should use the `Iterator` trait.
This would allow to fetch data from github as the iterator moves instead of all at once.

### LINKED ISSUE #20: use references when possible
in order to improve performances and avoid cloning String everywhere, we should use references whenever possible.

### LINKED ISSUE #21: consolidate errors in Logger API
Today, when dealing with multiple items to log, the logger will fail as soon as 1 item fails.
This is not convenient as we might want to upload a batch of PR and some will revert, some will succeed.
We should consolidate the errors instead of stopping at first failure.

AnthonyBuisset:
closes #19
closes #20
closes #21

### DIFF
--- a/src/bin/index_repos.rs
+++ b/src/bin/index_repos.rs
@@ -6,6 +6,7 @@ use deathnote_contributions_feeder::{
     database, github,
     model::pullrequest,
     model::repository,
+    services::contributions::fetch_and_log,
     traits::{fetcher::Fetcher, logger::Logger},
 };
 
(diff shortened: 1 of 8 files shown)


abdelhamidbakhta (approved):
lgtm

PR #32, opened 2022-07-04
bulk contributions upload for all registered users

AnthonyBuisset:
- :recycle: move Result out of Logger return type
- :sparkles: try to upload contribution only if user is registered
- :recycle: extract smart contract interactions from main logic
- :zap: bulk transactions when uploading contributions in smart contract

* Use the JSON RPC interface to perform a contract call to the Badge Registry to check if a given PR author is registered before trying to push its contribution.
* Contributions of all registered are pushed in batches to improve performances

Next step:
* Use the returned user information from this contract call to insert the contribution
* Wait for transaction to be accepted on L2 to send the next one

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -505,6 +505,7 @@ dependencies = [
  "serde_json",
  "starknet",
  "tokio",
+ "url",
  "uuid",
 ]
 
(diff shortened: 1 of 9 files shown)


abdelhamidbakhta (approved):
lgtm

PR #47, opened 2022-07-06
feat(database): add contribution_gates table, diesel structs, http api

tdelabro:


migrations/2022-07-05-171338_create_gates/up.sql
```diff
@@ -0,0 +1,6 @@
+CREATE TABLE contribution_gates(
+  id VARCHAR PRIMARY KEY,
```
ofux, line 2: I don't understand what this id is since there is already a gate_id and a contribution_id.

This table looks like a join table, and it seems weird to put an id in it.

src/routes/gates.rs
```diff
@@ -0,0 +1,53 @@
...
+
+    let database = database::API::new(connection);
+
+    database
+        .insert_gate(NewContributionGate {
+            contribution_id,
+            gate_id: body.gate_id.clone(),
+            transaction_hash: transaction_hash.clone(),
+        })
+        .map_err(|e| Failure::InternalServerError(e.to_string()))?;
```
ofux, line 50: To be done **after** starknetcc, but we'll have to think of a way to make this whole piece of code atomic, or different. Currently, the transaction can succeed and then the insert fails. In such case, we have a DB that is not in sync anymore with on-chain data.

AnthonyBuisset: we will move progressively into event-driven, so it won't be atomic because impossible to revert an on chain transaction once it has been sent, but it will be monitored and recoverable

migrations/2022-07-05-171338_create_gates/up.sql
AnthonyBuisset: we have our DB generated ID of a gate, and the on-chain generated ID (contract address I presume)
After starknet.cc we will need to brainstorm on how we should generate those Ids and which one will be the master one
As for example today we rely on github generated ids for projetcs and contributions

ofux: ok I see. Maybe the `gate_id` should be named `onchain_gate_id` or something like that then, but ok I understand now.
Indeed we'll have to think about our ids and how we manage them.

tdelabro: gate_id was the address of a on-chain gate contract.
So it would be exactly that: a join between a contribution and a gate contract

src/routes/gates.rs
tdelabro: yup. and the transaction can fail and the insert still happen

AnthonyBuisset (approved):

PR #49, opened 2022-07-06
feat(app): write skeleton of our application functions

tdelabro:


src/application.rs
```diff
@@ -0,0 +1,18 @@
+use anyhow::Result;
+
+use crate::domain::{Contribution, ContributorId};
+
+pub fn create_contribution(_contribution: Contribution) -> Result<()> {
+    Ok(())
+}
+
+pub fn attribute_contribution(
```
Bernardstanislas, line 9: maybe use `assign` instead of `attribute`, that's what we have in the ubiquitous language.
Ok I just added it :)

PR #55, opened 2022-07-07
:sparkles: add all contribution flows from latest smart contract ABI

AnthonyBuisset:


src/starknet/mod.rs
```diff
@@ -57,9 +57,9 @@ fn nb_transactions_in_batch() -> usize {
 
 pub fn oracle_contract_address() -> FieldElement {
```
tdelabro, line 58: we should probably stop calling it an oracle. It's just our contribution registry now

tdelabro (approved):

src/starknet/mod.rs
AnthonyBuisset: indeed, I will change that in a next PR to avoid conflicts

PR #59, opened 2022-07-08
✨ embed migrations in build

ofux:


abdelhamidbakhta (approved):
lgtm

src/database/mod.rs
```diff
@@ -27,6 +30,17 @@ pub fn establish_connection() -> Result<DbConn> {
         .map_err(anyhow::Error::msg)
 }
 
+fn establish_migration_connection() -> PgConnection {
```
AnthonyBuisset, line 33: there is already a function like this in connection.rs IIRC

Bernardstanislas (approved):

AnthonyBuisset (changes requested):

src/database/mod.rs
ofux: I tried to use it but it's a pool. The migration tool expects a PgConnection.
However I could call database_url() to avoid repeating the `env::var("DATABASE_URL").expect("DATABASE_URL must be set")`

PR #73, opened 2022-07-11
Add API feeder scripts

tekkac:
Adds a script to insert a list of contribution in the DB via the REST API.


scripts/feeder/contribution_feeder.py
```diff
@@ -0,0 +1,149 @@
...
+import csv
+from dotenv import load_dotenv
+
+
+CONTRIBUTIONS = "./scripts/feeder/contributions.csv"
+
+load_dotenv()
+API_KEY = os.environ["API_KEY"]
+GITHUB_KEY = os.environ["GITHUB_KEY"]
+HEADERS = {
```
ofux, line 13: looks like this constant is not used

scripts/feeder/contribution_feeder.py
```diff
@@ -0,0 +1,149 @@
...
+    "Accept": "application/json",
+    "Content-Type": "application/json",
+    "Api-Key": API_KEY
+}
+API_URL = "http://localhost:8000"
```
ofux, line 18:
```suggestion
API_URL = os.environ["API_URL"]
```

scripts/feeder/contribution_feeder.py
```diff
@@ -0,0 +1,149 @@
...
+    assert len(data) == 7
+    assert data[2] == 'github.com'
+    owner = data[3]
+    project = data[4]
+    issue = data[6]
+    return (owner, project, issue)
+
+GITHUB_API = "https://api.github.com/repos"
+
+def fetch_contribution_id(issue_url):
```
ofux, line 98: looks like this method is not used. Maybe simply remove it?

tekkac: yup, forgot to remove it from previous commit.

PR #75, opened 2022-07-12
:sparkles: fetch github data in GET /projects

### LINKED ISSUE #66: fetch github data during GET /projects
beware of github rate limiting
v0 alternative for #65

AnthonyBuisset:
closes #66

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -7,3 +7,4 @@ ACCOUNT_ADDRESS=0x0                                                             
 JSON_RPC_URI=https://13.38.227.15:9545                                                   # URI to the JSON RPC node
 NB_TRX_IN_BATCH=100                                                                      # Number of transactions sent in a batch
 API_KEY="ROOT"
+GITHUB_TOKEN="YOUR_PERSONAL_ACCESS_TOKEN"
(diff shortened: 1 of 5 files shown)


.env.example
```diff
@@ -7,3 +7,4 @@ ACCOUNT_ADDRESS=0x0
 JSON_RPC_URI=https://13.38.227.15:9545                                                   # URI to the JSON RPC node
 NB_TRX_IN_BATCH=100                                                                      # Number of transactions sent in a batch
 API_KEY="ROOT"
+GITHUB_TOKEN="YOUR_PERSONAL_ACCESS_TOKEN"
```
ofux, line 10: A personal access token in production? It's probably ok for the demo and starknetcc, but that's another thing we'll need to improve later.
I'm reading about Github Apps, which seems more adequate for this use case. However, in both cases, the rate limiting is going to be very problematic. 5000 per hour is extremely low.

ofux (approved):
LGTM, however I think we really have to tackle the rate limiting issue.
5000 req/hour is very low when you consider that every time we call /projects, we'll have has many request as issues. Let's say we have 100 issues, that means we can only make 50 calls to /projects per hour!

PR #97, opened 2022-07-19
perf(ProjectAPI): parallelize projects dto building in list_projects …

tdelabro:
…route

### DIFF
--- a/src/github/repo_cache.rs
+++ b/src/github/repo_cache.rs
@@ -2,6 +2,7 @@ use crate::utils::Cache;
 use octocrab::models::Repository;
 use std::{env, ops::Deref, time::Duration};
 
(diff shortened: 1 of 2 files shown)


tdelabro:
Note:
Using `try_join_all` the whole call will fail if one of the projects dto cannot be built.
We can also choose to use `join_all`, which returns a result for each task, and filter to only return dto that we were successfully able to build.
wdyt ?

AnthonyBuisset:
For the MVP I think it is better to have partial results rather than stopping at first failure. So I would choose the `join_all` option

Bernardstanislas (approved):

AnthonyBuisset (approved):

PR #105, opened 2022-07-25
feat(contributions): add logic to handle and store a contribution val…

tdelabro:
…idator


tdelabro:
I chose to have a nullable validator in DB, in order to be compliant with the contributions created before this update.

We can also choose to make it NON NULL and set the default value to "0x0" for the few that already exist. It will be more coherent with the future and simplify some parts of the code. I'm already doing this default thing, but on the fly at runtime instead of in the db.

wdyt?

AnthonyBuisset:
We are going to empty the DB one last time as we are deploying the proxy contract.
So we can have it NON NULL I guess

src/domain/entities/contribution.rs
```diff
@@ -1,7 +1,12 @@
-use crate::domain::*;
-use std::str::FromStr;
+use anyhow::anyhow;
+use starknet::core::types::FieldElement;
+use std::{fmt, str::FromStr};
 use url::Url;
 
+use super::{ContributorId, Project, ProjectId};
+
+pub type ValidatorAddress = FieldElement;
```
AnthonyBuisset, line 8: I think it is better to avoid talking about FieldElements as we decided (I think) that representation of addresses are going to be hexa strings

src/domain/entities/contribution.rs
```diff
@@ -1,7 +1,12 @@
-use crate::domain::*;
-use std::str::FromStr;
+use anyhow::anyhow;
```
Bernardstanislas, line 1: I thought we removed it from domain?

src/domain/entities/contribution.rs
```diff
@@ -34,19 +40,68 @@ pub struct Metadata {
 	pub r#type: Option<String>,
 }
 
+#[derive(Debug, PartialEq, Eq, Clone, Default)]
+pub struct Filter {
```
Bernardstanislas, line 44: Is it really business-related? I think it has been removed from @AnthonyBuisset refactor

src/infrastructure/database/project_repository.rs
```diff
@@ -95,6 +98,11 @@ impl From<models::Contribution> for Contribution {
 				context: contribution.context,
 				r#type: contribution.type_,
 			},
+			// ok to unwrap because values in db are created from FieldElement::ToString
+			validator: FieldElement::from_str(
+				&contribution.validator.unwrap_or_else(|| "0x0".to_string()),
```
Bernardstanislas, line 103: Agreed with @AnthonyBuisset , we can make it default in the DB, so that we remove this behaviour with a migration once the whole table has validators

src/routes/contributions.rs
```diff
@@ -14,30 +17,39 @@ use rocket::{
...
 
 use crate::action_queue::ActionQueue;
 
 use super::ApiKey;
 
 #[derive(Deserialize, JsonSchema)]
 #[serde(crate = "rocket::serde")]
-pub struct CreateContributionFromGitHubBody {
-	issue_number: u128,
+pub struct CreateContributionBody {
```
Bernardstanislas, line 28: `CreateContributionDto`

src/routes/contributions.rs
```diff
@@ -14,30 +17,39 @@ use rocket::{
...
 	_api_key: ApiKey,
-	body: Json<CreateContributionFromGitHubBody>,
+	body: Json<CreateContributionBody>,
 	github_api: &State<github::API>,
 	queue: &State<Arc<RwLock<ActionQueue>>>,
 ) -> Result<Status, Json<HttpApiProblem>> {
 	let body = body.into_inner();
-
-	let github_issue = github_api.issue(body.project_id, body.issue_number).await;
+	let validator = FieldElement::from_str(&body.validator).map_err(|e| {
```
Bernardstanislas, line 44: I know little about Rust, but can't this check be done somewhere inside the `CreateContributionDto` struct? It is an integrity check that should be held by the struct itself

Bernardstanislas (changes requested):

src/domain/entities/contribution.rs
```diff
@@ -34,19 +40,68 @@ pub struct Metadata {
...
+			Status::Assigned => write!(f, "ASSIGNED"),
+			Status::Completed => write!(f, "COMPLETED"),
+			Status::Abandoned => write!(f, "ABANDONED"),
+		}
+	}
+}
+
 impl FromStr for Status {
-	type Err = Error;
+	type Err = anyhow::Error;
```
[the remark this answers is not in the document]
tdelabro: Hey it's because I just delete this file and moved the one that was under `types/` in place. There was code duplication.

src/routes/contributions.rs
tdelabro: we probably can if we use some custom Deserialize implementation

src/domain/entities/contribution.rs
tdelabro: removed

src/domain/entities/contribution.rs
tdelabro: removed

src/infrastructure/database/project_repository.rs
tdelabro: done

src/routes/contributions.rs
tdelabro: done

src/domain/entities/contribution.rs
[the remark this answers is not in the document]
tdelabro: fixed

src/domain/entities/contribution.rs
AnthonyBuisset: the fact that we have a felt is tightly coupled with our usage of Starknet
If we want to consider other ecosystems, we should not have this kind of coupling in the code.
But I guess it is easy enough to refactor later when the need comes

src/routes/contributions.rs
Bernardstanislas: ok we'll do later then

src/domain/entities/contribution.rs
Bernardstanislas: Agreed with you @AnthonyBuisset, let's keep the `FieldElement` for now but we'll have to abstract it for further ecosystems integrations

src/routes/contributions.rs
tdelabro: done, and tested

src/infrastructure/database/contribution_repository.rs
```diff
@@ -70,6 +70,7 @@ impl From<(Contribution, String)> for models::NewContribution {
 			duration: contribution.metadata.duration,
 			context: contribution.metadata.context,
 			type_: contribution.metadata.r#type,
+			validator: contribution.validator.to_string(),
```
tekkac, line 73: `to_string` isn't `0x` prefixed. It is stored as a decimal.

Bernardstanislas (approved):
Ok for me once the CI is green :)

AnthonyBuisset (approved):

src/infrastructure/database/contribution_repository.rs
tdelabro: fixed

PR #107, opened 2022-07-26
WIP - Indexing of Starknet events based on apibara

AnthonyBuisset:
- :sparkles: listen to starknet events

This is a WIP
Implementation based on (apibara)[https://www.apibara.com/].

Comments are welcome, especially on the DDD part which can be improved.


Bernardstanislas (commented):
I suggest we put this PR as draft until we do the workshop that will decide what our strategy is, Apibara is promising for now but it's a huge bet

PR #108, opened 2022-07-27
🔨 update feeder script for latest API changes

ofux:


tekkac (approved):
Good catch
lgtm :muscle:

PR #109, opened 2022-07-29
refactor: add uuid to contribution

tdelabro:


src/domain/actions.rs
```diff
@@ -1,18 +1,20 @@
-use crate::domain::{Contribution, ContributionId, ContributorId};
+use crate::domain::*;
+
+use crate::domain::{Contribution, ContributorId};
```
AnthonyBuisset, line 3: duplicate with above line ?

src/domain/entities/contribution.rs
```diff
@@ -16,11 +17,12 @@ pub enum Status {
 	Abandoned = 3,
 }
 
-pub type Id = String;
+pub type OnChainId = String;
```
AnthonyBuisset, line 20: I think it is better to keep the alias for the id as well
to reference ContributionId or ContributionUuid outside of the module instead of Uuid
A type wrapper would be even better, it would prevent mistakes like passing a contributionId where a contributorId is expected for instance

src/domain/services/contribution.rs
```diff
@@ -1,12 +1,14 @@
+use uuid::Uuid;
+
 use crate::domain::*;
 
 pub trait Service {
 	fn create(&self, contribution: Contribution) -> Result<()>;
 	fn assign_contributor(
 		&self,
-		contribution_id: ContributionId,
+		contribution_id: Uuid,
```
AnthonyBuisset, line 9: for instance here it would be nicer to read ContributionId. This way, no leak of the implementation choice

src/infrastructure/database/contribution_repository.rs
```diff
@@ -1,53 +1,55 @@
...
+use crate::{domain::*, infrastructure::database::schema};
 use diesel::prelude::*;
 
 impl ContributionRepository for Client {
 	fn store(&self, contribution: Contribution, transaction_hash: String) -> Result<()> {
 		let contribution = models::NewContribution::from((contribution, transaction_hash));
 		diesel::insert_into(contributions::table)
 			.values(&contribution)
-			.get_result::<models::Contribution>(self.connection())
+			.execute(self.connection())
```
AnthonyBuisset, line 10: Nice !

src/infrastructure/database/models/contributions.rs
```diff
@@ -21,36 +22,38 @@ pub struct Contribution {
...
 
-#[derive(AsChangeset, Identifiable)]
-#[table_name = "contributions"]
-pub struct AssignContributionForm {
-	pub id: String,
-	pub status: String,
-	pub contributor_id: String,
-	pub transaction_hash: String,
-}
+// #[derive(AsChangeset, Identifiable)]
```
AnthonyBuisset, line 28: Can you please delete commented code ?

AnthonyBuisset (changes requested):

src/infrastructure/database/models/contributions.rs
Bernardstanislas, line 28: Why keeping this?

src/routes/contributions.rs
```diff
@@ -52,18 +53,20 @@ pub async fn create_contribution(
...
+			))
+		},
 	};
 
 	let metadata = github::extract_metadata(github_issue.clone());
 
 	let contribution = domain::Contribution {
-		id: (body.project_id * 1_000_000 + body.github_issue_number).to_string(),
+		id: Uuid::new_v4(),
+		onchain_id: (body.project_id * 1_000_000 + body.github_issue_number).to_string(),
```
Bernardstanislas, line 69: Don't we use uuids for onchain ids too? Or do we wait for the smart contract to be ready to handle this new type?

src/domain/entities/contribution.rs
```diff
@@ -16,11 +17,12 @@ pub enum Status {
 	Abandoned = 3,
 }
 
-pub type Id = String;
```
Bernardstanislas, line 19: I think we can still have the `Id` type, used as `ContributionId`, that equals a `Uuid`. So that we keep on talking about `ContributionId` everywhere else

src/utils/caches/issue_cache.rs
```diff
@@ -1,11 +1,11 @@
-use crate::domain::ContributionId;
 use octocrab::models::issues::Issue;
 use std::{env, ops::Deref, time::Duration};
+use uuid::Uuid;
 
 use super::Cache;
 
 #[derive(Clone)]
-pub struct IssueCache(Cache<ContributionId, Issue>);
+pub struct IssueCache(Cache<Uuid, Issue>);
```
Bernardstanislas, line 8: Regarding the remark on the `ContributionId`, I consider it can still exist next to the `OnChainContributionId`

src/domain/actions.rs
tdelabro: done

src/routes/contributions.rs
tdelabro: I don't know.
This code state does not break the front or the smart contracts. But it set the ground for later changes.
They can be done in another PR, when the time comes

src/domain/entities/contribution.rs
tdelabro: done

AnthonyBuisset (approved):

PR #110, opened 2022-07-29
Split CI workflow into unit jobs

Bernardstanislas:


.github/workflows/tests.yml
```diff
@@ -0,0 +1,41 @@
...
+  CARGO_TERM_COLOR: always
+
+jobs:
+  format:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - name: add nightly toolchain
+        run: rustup update nightly
+      - name: Cehck code format
```
AnthonyBuisset, line 16: typo

.github/workflows/tests.yml
```diff
@@ -0,0 +1,41 @@
...
+        run: rustup update nightly
+      - name: Check code linting
+        run: rustup component add clippy --toolchain nightly-x86_64-unknown-linux-gnu && cargo +nightly clippy --bins --tests --examples -- -D warnings
+  test:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - name: add nightly toolchain
+        run: rustup update nightly
+      - name: Run tests
```
AnthonyBuisset, line 40: Each step will re-build, is there a way to cache the build result ?

AnthonyBuisset (approved):

.github/workflows/tests.yml
Bernardstanislas: The job will run in parallel, there is no cache mechanism for inter-job caching (you can only pass explicit data from one job to another if you make them dependent).
I'm looking at this though https://actions-rs.github.io/ and I'll refactor :)

PR #111, opened 2022-07-29
Format logs as JSON

Bernardstanislas:


Cargo.toml
```diff
@@ -1,6 +1,7 @@
 [package]
 authors = ["Abdelhamid Bakhta <abdelhamid.bakhta@gmail.com>"]
 categories = ["contribution", "health-check", "analytics-tracking", "analytics"]
+default-run = "deathnote_contributions_feeder"
```
AnthonyBuisset, line 4: Shouldn't we rename to marketplace?

src/main.rs
```diff
@@ -36,7 +37,14 @@ embed_migrations!("migrations");
 #[tokio::main]
 async fn main() {
 	dotenv().ok();
-	env_logger::init();
+	let drain = slog_async::Async::default(slog_envlogger::new(
```
AnthonyBuisset, line 40: Could we refactor this into a function to explain what it does ?

AnthonyBuisset (approved):
LGTM.
Left some minor comments

Cargo.toml
Bernardstanislas: Yes I tried to do it in this PR but it's too big to mix both subjects

Cargo.toml
```diff
@@ -39,29 +39,35 @@ octocrab = "0.16.0"
...
+okapi = {version = "0.7.0-rc.1"}
+rocket = {version = "0.5.0-rc.2", features = ["json"]}
+rocket_okapi = {version = "0.8.0-rc.2", features = ["swagger"]}
 schemars = "0.8"
 
 # Starknet
-starknet = { git = "https://github.com/ofux/starknet-rs" }
-crypto-bigint = { version = "0.4.8", features = ["serde"] }
+crypto-bigint = {version = "0.4.8", features = ["serde"]}
+starknet = {git = "https://github.com/ofux/starknet-rs"}
```
tdelabro, line 50: I think olivier PR was merged into the main repo. We can go back to it

Bernardstanislas: They've merged it but I'm not sure they've created a new release for it https://github.com/xJonathanLEI/starknet-rs/tags

src/main.rs
Bernardstanislas: updated

Cargo.toml
ofux: They have indeed merged it, and we point to "https://github.com/ofux/starknet-rs" which means we take the last commit, not the last release :) So yeah, we can switch back to the main repo.
Sooner or later, we'll have to use releases.

PR #112, opened 2022-07-31
feat(routes): apply to contribution

tdelabro:


src/domain/repositories/uuid.rs
```diff
@@ -0,0 +1,5 @@
+use uuid::Uuid;
+
+pub trait Repository {
```
Bernardstanislas, line 3: I think this service is not a repository per se https://martinfowler.com/eaaCatalog/repository.html
I would name it `UuidGenerator` for example

src/infrastructure/uuid_generator.rs
```diff
@@ -0,0 +1,10 @@
+use crate::domain::*;
+use uuid::Uuid;
+
+pub struct UuidGenerator;
+
+impl UuidRepository for UuidGenerator {
```
Bernardstanislas, line 6: I feel this is so core to our models that it can be acceptable to have this directly implemented in the domain

src/routes/contributions/apply.rs
```diff
@@ -0,0 +1,44 @@
...
+	data = "<body>"
+)]
+pub async fn apply_to_contribution(
+	_api_key: ApiKey,
+	contribution_id: UuidParam,
+	body: Json<ApplyDto>,
+	connection: database::Connection,
+) -> Result<status::Accepted<()>, Json<HttpApiProblem>> {
+	let body = body.into_inner();
+	let uuid_generator = UuidGenerator;
```
Bernardstanislas, line 32: Can't it be injected by rocket?

src/routes/dto/to_http_api_problem.rs
```diff
@@ -0,0 +1,31 @@
+use deathnote_contributions_feeder::domain::Error;
+use http_api_problem::{HttpApiProblem, StatusCode};
+
+pub(crate) trait ToHttpApiProblem {
+	fn to_http_api_problem(&self) -> HttpApiProblem;
+}
+
+impl ToHttpApiProblem for Error {
```
AnthonyBuisset, line 8: why not implementing the From trait here ?

src/domain/entities/application.rs
```diff
@@ -0,0 +1,30 @@
+use super::{ContributionId, ContributorId};
+use uuid::Uuid;
+
+pub struct Application {
+	id: Uuid,
```
AnthonyBuisset, line 5: can we keep the same pattern here as well and create a type alias :
```
type Id = Uuid;
```

src/routes/contributions/apply.rs
tdelabro: Why should it be?

We are generating those `Uuid` purely randomly. There is no need for them to all come from the same single instance of `UuidGenerator`.
So need to make it a thread-safe managed state. Actually, there is no state to manage in `UuidGenerator`, it doesn't contain any data.

KISS, don't create a thread-safe managed state for something that is stateless

src/domain/repositories/uuid.rs
tdelabro: Is it a service? You just said `this service`

src/infrastructure/uuid_generator.rs
tdelabro: If we put it directly onto the domain, there is no need for a trait as we can use the domain struct anywhere.
On the other hand, we said want to be able to mock it. So we need a trait.
And pure domain objects are not supposed to be mockable, are they? That why we use this trait pattern.

src/infrastructure/database/models/applications.rs
```diff
@@ -0,0 +1,19 @@
+use crate::infrastructure::database::schema::*;
+use rocket::serde::{Deserialize, Serialize};
+use uuid::Uuid;
+
+#[derive(Identifiable, Queryable, Debug, Serialize, Deserialize)]
+#[serde(crate = "rocket::serde")]
+pub struct Application {
```
AnthonyBuisset, line 7: shouldn't we merge both structs as they are the same ?

AnthonyBuisset (changes requested):

src/infrastructure/database/repositories/application.rs
```diff
@@ -0,0 +1,28 @@
...
+	infrastructure::database::{models, schema::applications, Client},
+};
+
+impl ApplicationRepository for Client {
+	fn store(&self, application: Application) -> Result<()> {
+		let application = models::NewApplication::from(application);
+		diesel::insert_into(applications::table)
+			.values(&application)
+			.execute(self.connection())
+			.map_err(|e| Error::ApplicationStoreError(e.to_string()))?;
```
ltoussaint, line 14: Problem with sending this and the error management is that we can't make a difference between a internal sql error or the fact that the contributor already applied to the contribution (duplicate primary key)
I think it could be good to make the difference

src/routes/dto/to_http_api_problem.rs
tdelabro: Neither Error nor From are defined in this crate.
So I would have to create a wrapping type for the sole purpose of Implementing `From` on it.
I feel creating a new trait is more flexible. I can now implement this trait on anything I want, without having to wrap the type each time

src/domain/entities/application.rs
tdelabro: Sure

src/infrastructure/database/repositories/application.rs
tdelabro: I'm working on it, but it will be in another PR

src/infrastructure/database/models/applications.rs
tdelabro: Probably yes

tdelabro: But from what I remember it's a good pattern to keep Insertable and Queryable objects separated.
It just happens that they have the same field right now. But as it is two entirely different use-cases, it's better to keep them distincts.

tdelabro: Here is a person from the project explaining this: https://github.com/diesel-rs/diesel/issues/1440#issuecomment-354573185

src/domain/repositories/uuid.rs
tdelabro: I made it a service

src/infrastructure/uuid_generator.rs
tdelabro: I kept it in the infra for now

src/domain/repositories/uuid.rs
Bernardstanislas: Yes, a service for me is a quite generic name, a generator is more specific

src/infrastructure/uuid_generator.rs
Bernardstanislas: ok great you're right I forgot the mocking part

src/routes/contributions/apply.rs
Bernardstanislas: I think dependency injection will help with unit testing this route handler, regardless of threads considerations. But this can be refactored when we add tests. BTW wen tests?

src/routes/dto/to_http_api_problem.rs
AnthonyBuisset: Makes sense, thanks !

src/application/tests/apply_to_contribution.rs
```diff
@@ -0,0 +1,35 @@
...
+use uuid::Uuid;
+
+use crate::application::apply_to_contribution;
+
+use super::{InMemoryApplicationRepository, SerialUuidGenerator};
+
+#[test]
+fn store_and_find_application() {
+	let mut uuid_generator = SerialUuidGenerator(0);
+	let mut application_repository = InMemoryApplicationRepository::new();
```
AnthonyBuisset, line 12: `InMemoryApplicationRepository` would be something we create waiting for a DB to be created
For mocking purposes I would use `mockall` crate so that we can expect on calls
There is an example here: https://github.com/onlydustxyz/marketplace-backend/blob/main/src/routes/contributors/find.rs#L53-L58
This would also avoid adding the `mut`
Even if for a store (command) the `mut` can be expected

Bernardstanislas (approved):

src/domain/repositories/application.rs
```diff
@@ -1,5 +1,6 @@
 use crate::domain::*;
 
 pub trait Repository {
-	fn store(&self, application: Application) -> Result<()>;
+	fn store(&mut self, application: Application) -> Result<()>;
+	fn find(&self, id: &ApplicationId) -> Result<Application>;
```
AnthonyBuisset, line 5: we should have consistent names in all our repositories 
so far we have `ContributorRepository::by_id` and `ApplicationRepository::find`
WDYT ?

tdelabro: either `find` or `find_by_id` when there is multiple finding method

src/application/tests/apply_to_contribution.rs
tdelabro: The whole purpose of having a trait is that we don't care about what it is implemented on.
Here we are solely testing the `apply_to_contribution` application function. Whatever the repository is actually implemented on. This is not less or more valuable that using the Database implem and mocking the calls.

AnthonyBuisset: of course, I meant mocking the trait itself (the repository) instead of doing the inmemory implementation
but I agree both methods works for the purpose of unit testing the function

src/domain/repositories/application.rs
Bernardstanislas: I like the `find_by_{}` naming rule, pretty clear

PR #113, opened 2022-08-01
:sparkles: Find contributor by id (dummy implem)

AnthonyBuisset:
https://www.notion.so/onlydust/AAContributor-on-the-Swagger-when-I-call-the-GET-contributors-id-providing-my-contributor-id-I-59cddf2008fc4f3dba7342f6627d6ca4

### DIFF
--- a/src/application/mod.rs
+++ b/src/application/mod.rs
@@ -1,0 +1,2 @@
+mod get_contributor;
+pub use get_contributor::{GetContributor, Usecase as GetContributorUsecase};
--- a/src/routes/contributors/find.rs
+++ b/src/routes/contributors/find.rs
@@ -1,0 +1,110 @@
+use super::dto;
+use deathnote_contributions_feeder::application::{GetContributor, GetContributorUsecase};
+use http_api_problem::{HttpApiProblem, StatusCode};
+use rocket::{get, serde::json::Json};
+use rocket_okapi::openapi;
+use std::result::Result;
+
+#[openapi(tag = "Contributors")]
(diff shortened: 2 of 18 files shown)


src/application/mod.rs
```diff
@@ -0,0 +1,2 @@
+mod get_contributor_by_id;
+pub use get_contributor_by_id::{GetContributorById, Usecase as GetContributorByIdUsecase};
```
Bernardstanislas, line 2: I like when the usecases don't hold a technical detail in the name, what do you think about simply `GetContributor`

Bernardstanislas (approved):

src/routes/contributors/find.rs
```diff
@@ -0,0 +1,110 @@
...
+#[openapi(tag = "Contributors")]
+#[get("/contributors/<contributor_id>")]
+pub fn find_by_id(
+	usecase: GetContributorById,
+	contributor_id: u128,
+) -> Result<Json<dto::Contributor>, Json<HttpApiProblem>> {
+	find_by_id_impl(usecase, contributor_id)
+}
+
+fn find_by_id_impl<U: GetContributorByIdUsecase>(
```
ofux, line 17: It's not ideal, but it's not that bad. I think we can go this way. At least the application, domain, and infrastructure parts stay clean thanks to that.

ofux (approved):

src/application/mod.rs
AnthonyBuisset: I see no objection right

PR #114, opened 2022-08-01
:recycle: database client manages its own connection pool

AnthonyBuisset:


src/routes/projects/mod.rs
```diff
@@ -45,12 +44,10 @@ pub async fn new_project(
 #[openapi(tag = "Projects")]
 #[get("/projects")]
 pub async fn list_projects(
-	connection: database::Connection,
+	database: &State<database::Client>,
 	repo_cache: &State<caches::RepoCache>,
 	contributor_cache: &State<caches::ContributorCache>,
 ) -> Result<Json<Vec<dto::Project>>, Json<HttpApiProblem>> {
-	let database = database::Client::new(connection);
```
ofux, line 52: ❤️  no more internal infra inside HTTP handlers \o/

src/application/get_contributor.rs
```diff
@@ -2,34 +2,47 @@ use crate::{domain::*, infrastructure::database};
...
-impl Usecase for GetContributor {
+impl Usecase for GetContributor<'_> {
 	fn execute(&self, contributor_id: ContributorId) -> Result<Option<Contributor>> {
 		self.contributor_repository.by_id(contributor_id)
 	}
 }
 
 #[rocket::async_trait]
-impl<'r> FromRequest<'r> for GetContributor {
+impl<'r> FromRequest<'r> for GetContributor<'r> {
```
ofux, line 27: I know we didn't find a better way to do this, but I'm not really satisfied by this way of handling this dependency injection. I don't like the fact that the `GetContributor` object knows how it is going to be used in a HTTP handler. It breaks the Single Responsibility principle.
A proper dependency injection framework would hide this by doing it automatically.
I think I would even prefer doing this manually in the `main` function... but if I understood well, it is not possible because of the lifetime of managed stuff? Am I right?

ofux (approved):

src/application/get_contributor.rs
AnthonyBuisset: yes, it is much more difficult. But a real issue here I agree.
Let's try to tackle this in a dedicated refactoring, maybe PP ?

Bernardstanislas (approved):

src/application/mod.rs
```diff
@@ -2,13 +2,12 @@
 mod tests;
 
 mod get_contributor;
-
 pub use get_contributor::{GetContributor, Usecase as GetContributorUsecase};
 
 use crate::domain::*;
 
 pub fn apply_to_contribution<A: ApplicationRepository, U: UuidGenerator>(
-	application_repository: &mut A,
```
AnthonyBuisset, line 11: @tdelabro As the database client is now injected as a rocket::State, it cannot (unless I am wrong) be borrowed as mut
So I had to change it to a non mutable reference, and use `RefCell` in your `InMemory` implementation.
Is it ok for you ?

tdelabro: Nice. Let go with this.
The ref is not mutable itself, be we have interior mutability. Look good to me

PR #116, opened 2022-08-01
:ambulance: Fix unmanaged database state

AnthonyBuisset:


tdelabro (approved):
Why was it still compiling without it?

AnthonyBuisset:
> Why was it still compiling without it?

It seems there is no compile time error for this, only error log at run time 🤔

PR #124, opened 2022-08-03
fix: routes now return correct errors codes

tdelabro:
refactor: use struct based usecase for apply


src/routes/contributions/apply/tests.rs
```diff
@@ -0,0 +1,135 @@
+use std::{collections::HashMap, str::FromStr, sync::RwLock};
```
Bernardstanislas, line 1: I like the idea of this test, but its scope is a bit strange to me, it is between an integration test testing the API route with the real rocket library, but with a dummy usecase implementation...
It is ok for now since we don't have other tests on this part, but we might want to see either unit tests, or real integration tests in the future

Bernardstanislas (approved):

src/routes/contributions/apply/tests.rs
tdelabro: Yep that's what it is.
The use-case is already tested in the `application` module. Here we test the route (it's code and all rocket hidden one), using an injected dummy use-case.

PR #128, opened 2022-08-04
fix: make ContributionId its own type

tdelabro:


src/routes/contributions/create.rs
```diff
@@ -47,7 +47,7 @@ pub async fn create_contribution(
 	let metadata = github::extract_metadata(github_issue.clone());
 
 	let contribution = Contribution {
-		id: Uuid::new_v4(),
+		id: Uuid::new_v4().into(),
 		onchain_id: (body.project_id * 1_000_000 + body.github_issue_number).to_string(),
```
ofux, line 51: LGTM, but I thought we were setting `onchain_id` with the same value as `id` from now on. So I guess this is another story planed for later?

ofux (approved):

PR #131, opened 2022-08-04
feat: add list_applications route

tdelabro:


src/domain/entities/contributor.rs
```diff
@@ -38,3 +42,23 @@ impl From<u128> for Id {
 		Self(U256::from_u128(id))
 	}
 }
+
+impl JsonSchema for Id {
```
Bernardstanislas, line 46: For some reason this seems weird to me to have this in the domain. This is not core business. I think it should live in a presentation layer

src/infrastructure/database/repositories/application.rs
```diff
@@ -37,6 +37,32 @@ impl ApplicationRepository for Client {
 			res.map(|a| Some(a.into())).map_err_into()
 		}
 	}
+
+	fn list_for_contribution(
```
Bernardstanislas, line 41: `list_by_contribution` would be more consistent with the `find_by_` pattern we'd like to have in the repositories

src/routes/dto/hex_prefixed_string/mod.rs
```diff
@@ -1,23 +1,66 @@
...
+		&self.0[2..]
+	}
 }
 
 struct HexPrefixedStringVisitor;
 
+#[derive(Debug, Error)]
+pub enum ParseHexPrefixedStringError {
+	#[error("provided sting shoud be at least tree characters long")]
+	ToShort,
```
Bernardstanislas, line 34: `TooShort`

src/domain/entities/contributor.rs
tdelabro: Yup, but due to the orphan rule, It would force us to have a wrapper around each one of those types:
- ApplicationWrapper
- ApplicationIdWrapper
- ContributorIdWrapper
- ContributionIdWrapper
and so on

You just remark this one because I had to implement it myself. For every other one, it's done through a derive call.

src/routes/dto/hex_prefixed_string/mod.rs
tdelabro: done

src/infrastructure/database/repositories/application.rs
tdelabro: done

src/domain/entities/contributor.rs
tdelabro: I understand it breaks a principle. But it would be way simpler for everybody not to add all those wrapper boilerplate types.
Wdyt?

Bernardstanislas: I think that's a good reason not to follow the rule :)

Bernardstanislas (approved):

PR #136, opened 2022-08-05
:construction_worker: add full end-to-end test with starknet-devnet and create contribution

AnthonyBuisset:
- starknet-devnet run from a docker image
- load a previously prepared dump file at startup with contracts deployed
- documentation updated
- end-to-end test added that add a contribution, wait for the transaction to be executed and check the result


tekkac (approved):
Lgtm :muscle: :tada: 
The long-awaited e2e-testing is coming! :yellow_heart:

PR #141, opened 2022-08-09
test: make .env file sourcable

tdelabro:
allow to use `source .env.example`


tekkac (approved):
Good idea. Works great!

Bernardstanislas:
I am not sure this is a good practice, the standard for env files is to be only env vars, not shell commands. Won't work in all shells, see https://stackoverflow.com/questions/52232564/difference-between-export-some-env-var-key-and-someother-env-var-key-in-dotenv
We should use dotenv, this is the widely used best practice

PR #142, opened 2022-08-09
fix(API): assign contribution_id is of type ContributionId

tdelabro:


src/dto/u256.rs
```diff
@@ -0,0 +1,59 @@
+use crypto_bigint::{Encoding, U256};
+use thiserror::Error;
+
+use super::{hex_prefixed_from_str, ParseHexPrefixedStringError};
+
+#[derive(Debug, Error)]
+pub enum ParseU256Error {
+	#[error("provided sting shoud be between 3 and 66 characters long")]
+	InvalidLenght,
```
tekkac, line 9:
```suggestion
	InvalidLength,
```

src/routes/dto/u256/tests.rs
```diff
@@ -0,0 +1,30 @@
+use super::*;
+use serde_test::{assert_de_tokens, assert_de_tokens_error, Token};
+
+#[test]
+fn deserialize() {
+	assert_de_tokens(
+		&U256Param(U256::from_u128(81985529216486895)),
+		&[Token::Str("0x0123456789abcdef")],
```
tekkac, line 8: Does this mean we don't need to pad zeros to `contributor_id` anymore?

tekkac (approved):
that was a long read 
Lgtm

src/routes/dto/u256/tests.rs
tdelabro: yup id does. Took me a while to implement correctly but we are fine now

src/dto/u256.rs
tdelabro: done

PR #144, opened 2022-08-11
chore(e2e): make e2e a valid bin executable

tdelabro:
THis is how files should be organized in the cargo `bin` directory


tekkac (approved):
👌

PR #145, opened 2022-08-11
test(application): test error case for apply and accept_application

tdelabro:


src/application/contribution/accept_application.rs
```diff
@@ -14,40 +11,71 @@ pub trait Usecase: Send + Sync {
...
+			application_service,
 			application_repository,
 		})
 	}
 }
 
 impl Usecase for AcceptApplication {
 	fn accept_application(&self, application_id: &ApplicationId) -> Result<(), DomainError> {
-		let accepted_application = self
+		let application_to_accept = self
```
Bernardstanislas, line 52: This business logic should live in the domain, the application layer is here to coordinate calls to the domain services

src/application/contribution/accept_application.rs
```diff
@@ -14,40 +11,71 @@ pub trait Usecase: Send + Sync {
...
+			.ok_or_else(|| DomainError::from(ApplicationRepositoryError::NotFound))?;
+
+		if *application_to_accept.status() != ApplicationStatus::Pending {
+			return Err(DomainError::InvalidApplicationStatus {
+				current: *application_to_accept.status(),
+				required: ApplicationStatus::Pending,
+			});
+		}
+
+		self.application_service
```
Bernardstanislas, line 65: This should hold the logic you described above

src/application/contribution/accept_application.rs
```diff
@@ -14,41 +11,52 @@ pub trait Usecase: Send + Sync {
...
+			application_service,
 		})
 	}
 }
 
 impl Usecase for AcceptApplication {
 	fn accept_application(&self, application_id: &ApplicationId) -> Result<(), DomainError> {
-		let accepted_application = self
-			.application_repository
+		let application = self
```
Bernardstanislas, line 47: This is crystal clear :)

src/infrastructure/database/domain_implementations/application/service.rs
```diff
@@ -19,30 +19,79 @@ impl ApplicationService for Client {
 			.connection()
 			.map_err(|e| ApplicationServiceError::Infrastructure(Box::new(e)))?;
```
Bernardstanislas, line 20: This logic should be in the domain, not in the infrastructure. `ApplicationService`should be a struct, using an `ApplicationRepository` to fetch/save data, and having the business rules directly coded in its methods.
The infrastructure then only needs to implement the repository.

Bernardstanislas (changes requested):
Business logic must live in the domain, here the domain is only present in the form of an `ApplicationService` trait. This should be made a struct that holds the business rules (only accept a pending application,refuse all the others, only apply to an open contribution etc). This can also be done in a `ContributionService`, we would then need to distinguish it from our existing `ContributionService` trait that can be renamed to `OnChainContributionService`

Bernardstanislas (approved):

PR #146, opened 2022-08-11
CI: add a new job for tests requiring a db presence

tdelabro:


tekkac (approved):
Lgtm

Bernardstanislas (approved):
This is a good first step, next step according to me is to use "native" github action mechanisms such as service containers https://docs.github.com/en/actions/using-containerized-services/about-service-containers

PR #148, opened 2022-08-12
Refactor contribution service on apply usecase

Bernardstanislas:
This PR refactors the way the backend handles the apply business action.

I did it to show an example of how usecases must use the domain, keeping the business logic inside the domain layer.

This refactoring must be carried on with subsequent PRs to deal with all the other usecases.

It introduces a distinction between the onchain contribution service, and the contribution service (offchain is implied).


AnthonyBuisset (approved):
I am ok with this re-design

PR #153, opened 2022-08-19
refactor code base to use Cargo workspace feature and structure code into crates

AnthonyBuisset:
New workspace structure:
    - backend crate
    - domain crate
    - wrappers crate

Removal of dead code

- :recycle: initiate Cargo workspace
- :recycle: rename deathnote-contribution-feeder to marketplace-backend in Cargo.toml
- :recycle: rename folders and move domain into its own crate
- ♻️ rename DomainError as Error in domain crate
- 🔒️ fix clippy warnings
- :coffin: remove unused contract update status
- :coffin: remove unused types module
- :coffin: remove unused interface
- :heavy_minus_sign: remove dependency to anyhow in domain crate
- :coffin: remove unused interface
- :coffin: remove unused contract implementation and traits module
- :memo: update README in crates


Cargo.toml
```diff
@@ -1,100 +1,7 @@
...
-assert-json-diff = "2.0.2"
-
-# Local dependecies
-wrappers = { path = "./wrappers" }
-
-[dev-dependencies]
-serde_test = "1.0.140"
-assert_matches = "1.5"
+members = [
+    "marketplace-backend",
```
Bernardstanislas, line 4: is it possible to simply name it `marketplace`?

Cargo.toml
```diff
@@ -1,100 +1,7 @@
...
+    "wrappers",
+	"onlydust-domain"
```
Bernardstanislas, line 6: is it possible to simply name it `domain`?

Bernardstanislas (approved):
Candide questions :)

Cargo.toml
AnthonyBuisset: Yes it is possible, but then it does not allow having a domain inside a crate anymore...
Usually, the split is feature-based in other projects, for instance:
```
[workspace]
members = [
    "starknet-core",
    "starknet-providers",
    "starknet-contract",
    "starknet-crypto",
    "starknet-signers",
    "starknet-accounts",
    "starknet-ff",
    "starknet-macros",
    "examples/starknet-wasm",
]
```

Here, taking out the whole domain is feels weird...

Cargo.toml
AnthonyBuisset: should we rename the crate marketplace or the repository ?
as the mono-repo starts getting bigger and bigger

tdelabro: the marketplace actually span through our backend, our front and the on-chain logic. So it can be relevant to specify which part of it is it

Cargo.toml
tdelabro: `marketplace-domain` could also be a good one

Cargo.toml
AnthonyBuisset: how about:
```
members = [
    "marketplace-core",
    "marketplace-indexer",
    "marketplace-domain",
    "marketplace-utils",
]
```
@tdelabro @Bernardstanislas

marketplace-backend/src/infrastructure/database/domain_implementations/application/repository.rs
```diff
@@ -42,7 +40,7 @@ impl ApplicationRepository for Client {
 		if let Err(diesel::result::Error::NotFound) = res {
 			Ok(None)
 		} else {
-			res.map(|a| Some(a.into())).map_err_into()
+			res.map(|a| Some(a.into())).map_err(DatabaseError::from).map_err_into()
```
tdelabro, line 43: Why do you have to call `map_err()` and `map_err_into()` back to back.
Shouldn't `map_err_into` be enough?

marketplace-backend/src/infrastructure/database/domain_implementations/application/service.rs
```diff
@@ -54,21 +54,21 @@ impl ApplicationService for Client {
 				Ok(application)
 			});
 
-		res.map_into().map_err_into()
+		res.map_into().map_err(DatabaseError::from).map_err_into()
```
tdelabro, line 57: Same. Is there no way to go from DieselError to ApplicationServiceError directly?
Instead of doing DieselError -> DatabaseError -> ApplicationServiceError
Do this extra step and this extra type serve a purpose?

marketplace-backend/src/infrastructure/database/domain_implementations/application/repository.rs
AnthonyBuisset: because we need an extra step:
`diesel` -> `DtabaseError` -> `DomainError`

For a reason I don't know, this would work:
```
let res = res.map(|a| Some(a.into())).map_err(DatabaseError::from)?;
Ok(res)
```

It seems the `?` is calling `From`

marketplace-backend/src/infrastructure/database/domain_implementations/project.rs
```diff
@@ -46,20 +46,20 @@ impl ProjectRepository for Client {
...
-	}
-}
+// impl From<(models::Project, Vec<models::Contribution>)> for ProjectWithContributions {
+// 	fn from((project, contributions): (models::Project, Vec<models::Contribution>)) -> Self {
+// 		Self {
+// 			project: project.into(),
+// 			contributions: contributions.into_iter().map_into().collect(),
+// 		}
+// 	}
+// }
```
tdelabro, line 62: Should we remove it now?

marketplace-backend/src/infrastructure/database/domain_implementations/application/service.rs
AnthonyBuisset: yes, ApplicationServiceError is in another crate :(

marketplace-backend/src/infrastructure/database/domain_implementations/project.rs
AnthonyBuisset: yes indeed !

marketplace-backend/src/infrastructure/database/error.rs
```diff
@@ -6,4 +7,6 @@ pub enum Error {
 	Connection(String),
 	#[error("Unable to migrate the database")]
 	Migration(String),
+	#[error(transparent)]
+	Diesel(#[from] DieselError),
```
tdelabro, line 11: I'm not sure about this.
Connection and Migration are also Diesel errors in a way.

AnthonyBuisset: True... should we map into more specific errors or just use diesel error ?
As the domain error is specific anyway and DatabaseError is just used to forward from diesel to domain error

onlydust-domain/Cargo.toml
```diff
@@ -0,0 +1,49 @@
...
+uuid = { version = "0.8.2", default_features = false, features = [
+	"v4",
+	"serde",
+] }
+crypto-bigint = { version = "0.4.8", features = ["serde"] }
+url = "2.2.2"
+base16ct = "0.1.1"
+# TODO: Remove both dependencies below once https://www.notion.so/onlydust/remove-dependency-on-schema-rs-in-domain-crate-584da1e29638490986f2e28e3988709f has been done
+schemars = { version = "0.8", features = ["uuid1"] }
+# rocket = { version = "0.5.0-rc.2", features = ["json", "uuid"] }
```
tdelabro, line 27: remove commented dependecy

marketplace-backend/src/infrastructure/database/schema.rs
```diff
@@ -0,0 +1,52 @@
+table! {
```
tdelabro, line 1: A second schema file has been added.
You have to remove it and update this value in `diesel.toml` in order to specify the new path for this auto-generated file

```
[print_schema]
file = "src/infrastructure/database/schema.rs"
```

AnthonyBuisset: well spotted !

Cargo.toml
Bernardstanislas: amazing :) what is the `marketplace-utils` member? Seeing utils is a personal code smell :)

tdelabro (changes requested):
I added some questions and a change is required in in `diesel.toml`

marketplace-backend/src/infrastructure/database/domain_implementations/application/service.rs
tdelabro: Ohhhh that's why!
Well I guess that is the price we have to pay

marketplace-backend/src/infrastructure/database/error.rs
tdelabro: `Transaction` maybe or `Execution`, since `DieselError` only contains errors that happen when actually executing stuff in the database?

Cargo.toml
AnthonyBuisset: today it only includes wrappers, we can keep marketplace-wrappers as today

marketplace-backend/src/infrastructure/database/domain_implementations/application/repository.rs
tdelabro: Ok, probaly some compiler limitation. Or the way `map_err_into` is implemented

tdelabro (approved):

PR #155, opened 2022-08-19
:sparkles: create marketplace-indexer crate for interfacing with apibara

AnthonyBuisset:


marketplace-domain/src/value_objects/events/github_identifier_registered.rs
```diff
@@ -0,0 +1,39 @@
+use crate::*;
```
Bernardstanislas, line 1: Events are such a specific value object type that they can live in their own folder directly in the domain, like `marketplace-domain/src/events/github_identifier_registered.rs`

PR #156, opened 2022-08-22
feat: migrate contributor_id in contribution table to be 0x prefixed

tdelabro:


migrations/2022-08-22-135718_move-contributorid-to-0x-prefixed-format/up.sql
```diff
@@ -0,0 +1,2 @@
+UPDATE contributions
+    SET contributor_id = '0x' || contributor_id
```
ofux, line 2:
```suggestion
    SET contributor_id = '0x' || contributor_id
    WHERE contributor_id LIKE '000%'
```

migrations/2022-08-22-135718_move-contributorid-to-0x-prefixed-format/down.sql
```diff
@@ -0,0 +1,2 @@
+UPDATE contributions
+    SET contributor_id = right(contributor_id, -2)
```
ofux, line 2:
```suggestion
    SET contributor_id = right(contributor_id, -2)
    WHERE contributor_id LIKE '0x%'
```

ofux (changes requested):
I prefer to be as safe as possible when writing this kind of migrations ;)

migrations/2022-08-22-135718_move-contributorid-to-0x-prefixed-format/up.sql
tdelabro: what about `WHERE contributor_id NOT LIKE '0x%'` instead

ofux: yes, sounds good

ofux (approved):

PR #159, opened 2022-08-22
kick-off event sourcing migration

AnthonyBuisset:
- :recycle: move contribution status into its own file
- :recycle: use ContractAddress type for validator address in domain instead of felt
- :recycle: change HexPrefixedString implementation to make Event serializable
- ♻️ refactor event unit test to use fixture
- ♻️ use HexPrefixedString as ContributorId internal representation to have proper serde
- :sparkles: add ContributionEvent support
- :heavy_plus_sign: add eventually crate
- :sparkles: create Contribution aggregate
- :card_file_box: add events table
- :recycle: move contribution aggregate out of entity folder
- :sparkles: implement EventStore for Database
- :recycle: remove dependency on eventually and re-implement needed traits
- :sparkles: translate starknet events into contribution events
- :coffin: remove unused event
- :recycle: create HexStringWrapper and use it as Contribution aggregate ID
- :recycle: event should contain aggregate id


Bernardstanislas (commented):
First of all, THANK YOU for the atomic commits, this helped review in a great way.

Why did you end up not using eventually? I found it pretty clear tu use, but I'm also ok with having our own traits and implementations since it's not a big part.

From what I understood in the end of the PR, you still need to add actual event sourcing right? Meaning being able to create a `ContributionAggregate` from an event stream, updating projections etc.


The `Aggregate` trait shouldn't be located in the `events` folder IMHO, it is a basic building block in the domain, it should live right in `marketplace-domain/src`.

It is not in the scope of this PR but at some point you will introduce a `ContributionAggregateRepository`, needed to perform backend business actions such as `apply to contribution`. This repository will differ from the projections' ones because it will be fully implemented in the domain, getting injected the `EventStore` and iteration over the events calling the aggregate's `apply` – apply event, not apply to contribution. Beware that this repository will obviously not have any write method.

Final remark, I don't think it's a good idea to add business checks in the `apply` method of an aggregate. One must consider that once the event reached the event store, it is valid in the stream of events business-wise. In our case, this is ensured by the blockchain, and by the code we will add to handle contribution application that will make business checks before emitting an event.
Having a check at this point brings the risk of "bricking" a contribution, meaning not being able to recreate the aggregate from the event stream, and thus making any action on this contribution impossible. This is to avoid at any cost.

AnthonyBuisset:
> Why did you end up not using eventually? I found it pretty clear tu use, but I'm also ok with having our own traits and implementations since it's not a big part.

It required too many things to implement that were not needed, so it was quicker to create our own traits

AnthonyBuisset:
> From what I understood in the end of the PR, you still need to add actual event sourcing right? Meaning being able to create a `ContributionAggregate` from an event stream, updating projections etc.

Indeed, but the PR was already big enough :-)

AnthonyBuisset:
> The `Aggregate` trait shouldn't be located in the `events` folder IMHO, it is a basic building block in the domain, it should live right in `marketplace-domain/src`.

You're right, I will change that

AnthonyBuisset:
> It is not in the scope of this PR but at some point you will introduce a `ContributionAggregateRepository`, needed to perform backend business actions such as `apply to contribution`. This repository will differ from the projections' ones because it will be fully implemented in the domain, getting injected the `EventStore` and iteration over the events calling the aggregate's `apply` – apply event, not apply to contribution. Beware that this repository will obviously not have any write method.

Indeed, it will soon be needed, as well as the event bus trait

AnthonyBuisset:
> Final remark, I don't think it's a good idea to add business checks in the `apply` method of an aggregate. One must consider that once the event reached the event store, it is valid in the stream of events business-wise. In our case, this is ensured by the blockchain, and by the code we will add to handle contribution application that will make business checks before emitting an event.
> Having a check at this point brings the risk of "bricking" a contribution, meaning not being able to recreate the aggregate from the event stream, and thus making any action on this contribution impossible. This is to avoid at any cost.

Got it, I will remove the checks and turn the return value from `Result<Event>` to `Event`

PR #164, opened 2022-08-25
refactor: removed the need for shema crate in domain

tdelabro:
fix: contribution create does not take 'validator' any more, but it is still written in db


AnthonyBuisset (approved):
LGTM

PR #165, opened 2022-08-25
🎨 🐛 update application and contribution status when transaction is done

ofux:
Introduce the new usecase `on_contribution_assigned`

end2end tests will be added in another PR

### DIFF
--- a/marketplace-core/src/application/contribution/on_contribution_assigned.rs
+++ b/marketplace-core/src/application/contribution/on_contribution_assigned.rs
@@ -61,7 +61,8 @@ impl Usecase for OnContributionAssigned {
 		contribution.contributor_id = Some(contributor_id.to_owned());
 		self.contribution_repository.update(contribution).map_err(DomainError::from)?;
 
-		// Note that we handle the None case properly as the old way to apply (ie. typeform) didn't create any application.
+		// Note that we handle the None case properly as the old way to apply (ie. typeform) didn't
+		// create any application.
 		match self
 			.application_repository
 			.list_by_contribution(&contribution_id, Some(contributor_id.to_owned()))
--- a/marketplace-core/src/application/contribution/on_contribution_assigned.rs
+++ b/marketplace-core/src/application/contribution/on_contribution_assigned.rs
@@ -1,0 +1,231 @@
+use std::sync::Arc;
+
+use marketplace_domain::{Error as DomainError, *};
+
+pub trait Usecase: Send + Sync {
+	fn on_assigned(
+		&self,
+		contribution_onchain_id: ContributionOnChainId,
+		contributor_id: &ContributorId,
+	) -> Result<(), DomainError>;
+}
+
+pub struct OnContributionAssigned {
+	contribution_repository: Arc<dyn ContributionRepository>,
+	application_repository: Arc<dyn ApplicationRepository>,
+	application_service: Arc<dyn ApplicationService>,
+}
+
+impl OnContributionAssigned {
+	pub fn new(
+		contribution_repository: Arc<dyn ContributionRepository>,
+		application_repository: Arc<dyn ApplicationRepository>,
+		application_service: Arc<dyn ApplicationService>,
+	) -> Self {
+		Self {
+			contribution_repository,
+			application_repository,
+			application_service,
+		}
+	}
+}
+
+impl OnContributionAssigned {
+	pub fn new_usecase_boxed(
+		contribution_repository: Arc<dyn ContributionRepository>,
+		application_repository: Arc<dyn ApplicationRepository>,
+		application_service: Arc<dyn ApplicationService>,
+	) -> Box<dyn Usecase> {
+		Box::new(Self {
+			contribution_repository,
+			application_repository,
+			application_service,
+		})
+	}
+}
+
+impl Usecase for OnContributionAssigned {
+	fn on_assigned(
+		&self,
+		contribution_onchain_id: ContributionOnChainId,
+		contributor_id: &ContributorId,
+	) -> Result<(), DomainError> {
+		let mut contribution = self
+			.contribution_repository
+			.find_by_onchain_id(&contribution_onchain_id)
+			.map_err(DomainError::from)?
+			.ok_or_else(|| DomainError::from(ContributionRepositoryError::NotFound))?;
+
+		let contribution_id = contribution.id;
+		contribution.status = ContributionStatus::Assigned;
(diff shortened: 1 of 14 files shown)


marketplace-core/src/application/contribution/on_contribution_assigned.rs
```diff
@@ -0,0 +1,231 @@
...
+		contributor_id: &ContributorId,
+	) -> Result<(), DomainError> {
+		let mut contribution = self
+			.contribution_repository
+			.find_by_onchain_id(&contribution_onchain_id)
+			.map_err(DomainError::from)?
+			.ok_or_else(|| DomainError::from(ContributionRepositoryError::NotFound))?;
+
+		let contribution_id = contribution.id;
+		contribution.status = ContributionStatus::Assigned;
```
Bernardstanislas, line 60: All the logic in this usecase should be in the `ApplicationService`. This is not technical, this is a business rule, let's put it in the domain

marketplace-domain/src/services/application.rs
```diff
@@ -24,6 +25,8 @@ pub enum Error {
 	},
 }
 
+#[automock]
 pub trait Service: Send + Sync {
-	fn accept_application(&self, id: &ApplicationId) -> Result<Application, Error>;
+	fn accept_application(&self, application: Application) -> Result<(), Error>;
```
Bernardstanislas, line 30: Why is it still abstract? This service can be fully implemented, it relies on the `ApplicationRepository` and the `ContributionRepository` and calls the `update` methods.
Here all the logic is either in the use case, or in the infrastructure. Shouldn't be there, it is domain logic. 
See this PR https://github.com/onlydustxyz/marketplace-backend/pull/148 to have an example.
Moreover, reading this former PR, I would consider the `accept_application` method as part of the `ContributionService` rather than having an `ApplicationService` which doesn't do much and is always related to a contribution.

marketplace-infrastructure/src/database/domain_implementations/application/service.rs
```diff
@@ -10,51 +10,65 @@ use crate::database::{
...
+			.connection()
+			.map_err(|e| ApplicationServiceError::Infrastructure(Box::new(e)))?;
+
+		let res: Result<(), diesel::result::Error> = connection.transaction(|| {
+			// Set all applications to refused
+			diesel::update(
+				applications::dsl::applications
+					.filter(applications::contribution_id.eq(contribution_id.as_uuid())),
+			)
+			.set(applications::status.eq(Status::Refused))
```
Bernardstanislas, line 65: This is a consequence of having an anemic domain: very specific code in the infra layer, where it should only be an `update` method that updates the whole entity, it is not the role of the infra to know which part will get updated.
See https://github.com/onlydustxyz/marketplace-backend/blob/main/marketplace-infrastructure/src/database/domain_implementations/contribution.rs#L51

Bernardstanislas (changes requested):
You should avoid having business logic anywhere else than in the domain.
Based on the comments I wrote below, I would even consider removing the trait `ApplicationService`, since the application is never the entrypoint but the contribution is, and code the business logic in the `ContributionService`

marketplace-infrastructure/src/database/domain_implementations/application/service.rs
```diff
@@ -10,51 +10,65 @@ use crate::database::{
...
-		let res: Result<models::Application, diesel::result::Error> =
-			connection.transaction(|| {
-				// Set the chosen one to accepted
-				let application = diesel::update(
-					applications::dsl::applications.filter(applications::id.eq(id.as_uuid())),
-				)
-				.set(applications::status.eq(Status::Accepted))
-				.get_result::<models::Application>(&*connection)?;
+		let res: Result<(), diesel::result::Error> = connection.transaction(|| {
+			// Set the chosen one to accepted
```
AnthonyBuisset, line 26: typo

marketplace-domain/src/services/application.rs
AnthonyBuisset: Indeed, in Rust we can set a default implementation that could still be overridden for mocks.
Agreed to have this in the ContributionService instead, makes more sense

marketplace-domain/src/services/application.rs
```diff
@@ -24,6 +25,8 @@ pub enum Error {
 	},
 }
 
+#[automock]
```
AnthonyBuisset, line 28: can you please add automock in test only ? I will soon refactor the code base to have this new convention everywhere

marketplace-core/src/application/contribution/on_contribution_assigned.rs
```diff
@@ -0,0 +1,231 @@
...
+		};
+		let mut updated_contribution = contribution.clone();
+		updated_contribution.status = ContributionStatus::Assigned;
+		updated_contribution.contributor_id = Some(contributor_id.to_owned());
+
+		let returned_contribution = contribution.to_owned();
+		contribution_repository
+			.expect_find_by_onchain_id()
+			.with(eq(contribution_onchain_id.to_owned()))
+			.returning(move |_| Ok(Some(returned_contribution.to_owned())));
```
AnthonyBuisset, line 184: why do you need to call `to_owned` several times ?

marketplace-domain/src/services/application.rs
ofux: Seen with @AnthonyBuisset : will be done in a next refactoring

marketplace-domain/src/services/application.rs
ofux: I 100% agree with you. Currently, the repository literally implements the ApplicationService, which is a bad idea. However, I decided not to refactor this part in this PR as it would have needed at least 0.5 more days.
I suggest -as discussed with @AnthonyBuisset- that we do this refactoring in another Task.

marketplace-infrastructure/src/database/domain_implementations/application/service.rs
ofux: Agreed, see my previous comment

AnthonyBuisset (approved):
As discussed together, we will pair parogram to finish the refactoring as it conflicts with my current tasks

marketplace-core/src/application/contribution/on_contribution_assigned.rs
ofux: Agreed. As discussed with @AnthonyBuisset we will merge it as-is and pair-program on the next Task to avoid too many conflicts.

PR #167, opened 2022-08-26
feat: remove ActionQueue

tdelabro:
1)
I used tokio `block_on` method, to be able to call an async function inside a synchronous one.
It's not good at all: https://rust-lang.github.io/async-book/01_getting_started/04_async_await_primer.html
```
Unlike block_on, .await doesn't block the current thread, but instead asynchronously waits for the future to complete, allowing other tasks to run if the future is currently unable to make progress.
```

The solution would be to make usecases async. So everything is async, from the rocket route to the send transaction service. But because we do this weird thing of having usecases being actual instantiated objects managed by the rocket state, they have to be `send` and `sync`, which prohibits us from calling any form of async code inside (await or join).

So, either I missed something and we can actually find an another way to have async code executed inside a synchronous function, or we have to stop with this weird thing of having usecases being actual in-memory instances.
We did it for the sake of DDD but if it starts to hurt performances, I would refer to find another way.
Basically, just call the function itself instead of instantiating an object and sharing it between threads, just to finally call the method.

2)
All those starknet service methods now return the transaction hash. I did not do the extra step of returning it to the caller because it impact the frontend.
It will be another user story, with two tasks, for front and back ends


marketplace-infrastructure/src/starknet/contracts/utils/contract_administrator.rs
```diff
@@ -81,49 +74,4 @@ impl<A: Account + Sync> ContractAdministrator<A> {
 			})
 		}
 	}
-
-	pub async fn wait_for_transaction_acceptance(
```
ofux, line 85: feel so good to remove this one ❤️

ofux (approved):
LGTM

PR #171, opened 2022-08-28
Actual event sourcing

Bernardstanislas:
Implement event sourcing with the contribution aggregate root, and use it as a source of truth in a first use case – others are marked as todos.


marketplace-domain/src/contribution/repositories/contribution.rs
```diff
@@ -0,0 +1,109 @@
...
+	fn new(event_store: Arc<dyn EventStore<ContributionAggregateRoot>>) -> Self {
+		Self { event_store }
+	}
+}
+
+impl Repository for RepositoryImplementation {
+	fn find_by_id(&self, id: ContributionId) -> Result<ContributionAggregateRoot, Error> {
+		let events = self.event_store.list_by_id(&id)?;
+		if events.len() == 0 {
+			return Err(Error::NotFound);
```
AnthonyBuisset, line 34: I think the standard we chose (yet to be written) is to return a Return<Option<T>, Error> instead of Err::NotFound. Unless it should never happen ?

marketplace-domain/src/contribution/aggregate_root/mod.rs
```diff
@@ -7,55 +7,66 @@ use std::{fmt::Display, str::FromStr};
...
-
 mod status;
 pub use status::Status;
 
 #[derive(Clone, Debug, PartialEq, Eq, Serialize, Deserialize, Hash, Default, HexStringWrapper)]
 pub struct Id(HexPrefixedString);
 
-pub struct Aggregate;
+#[derive(Default, Debug, Clone, PartialEq, Eq)]
+pub struct Aggregate {
```
AnthonyBuisset, line 17: so an `Aggregate` is the previous `Aggregate::State` and the `AggregateRoot` trait is the previous `Aggregate` ?
An `Aggregate` no longer has State. 

It differs from `eventually-rs` crate implementation and what I have read about ES.

I think it is important to settle on a version and write a standard for it. I prepared one here:https://www.notion.so/onlydust/Event-driven-architecture-4aeb8e647a474735b5921c4370265544

marketplace-domain/src/aggregate_root.rs
```diff
@@ -1,7 +1,9 @@
 pub trait AggregateRoot: Send + Sync {
 	type Id;
-	type State;
 	type Event;
+}
 
-	fn apply(state: Self::State, event: Self::Event) -> Self::State;
+pub trait EventSourceable: AggregateRoot {
```
AnthonyBuisset, line 6: This means that only `AggregateRoot` can be event sourced, not a regular `Aggregate` ?

marketplace-domain/src/contribution/repositories/contribution.rs
Bernardstanislas: Actually I found a strange mix in [multiple](https://github.com/onlydustxyz/marketplace-backend/blob/main/marketplace-domain/src/repositories/contact_information.rs#L10) [places](https://github.com/onlydustxyz/marketplace-backend/blob/main/marketplace-domain/src/repositories/contribution_projection.rs#L9), having both the `NotFound` error and the `Option` does not make sense. I tend to prefer having the explicit error, hence my removal of the `Option`

marketplace-domain/src/contribution/aggregate_root/mod.rs
Bernardstanislas: Aggregate is the notion of a group of coherent events, which in a simplified conception can be viewed as a DDD bounded context.
The aggregate root is simply the aggregate you use to make any business action in the given bounded context.
In our case, both the aggregate and the aggregate root are the notion of contribution.

See https://stackoverflow.com/questions/41914040/event-store-root-aggregate-vs-aggregate for more info.

And for the state, we don't need to over model the aggregate root, it has its own fields and it works well this way.

marketplace-domain/src/aggregate_root.rs
Bernardstanislas: See my previous answer, both concepts are mixed and it's ok

marketplace-domain/src/contribution/repositories/contribution.rs
AnthonyBuisset: Opportunity for a standard?

marketplace-domain/src/aggregate_root.rs
AnthonyBuisset: I would like to counter propose a design, tell me if it makes sense or if I misunderstood something:

An aggregate is a group of events, can handle them and be sourced by them
```
pub trait Aggregate {
   fn handle_event(...)
   fn from_events(...)
}

```
An aggregate root is a special kind of aggregate: 
```
pub trait AggregateRoot: Aggregate {}
```

The state of an aggregate at a given time, concrete type:
```
pub struct State {}

impl Aggregate for State {}
// and if needed
impl AggregateRoot for State {}
```

And the repository that can managed only aggregate roots
```
pub trait Repository<AR: AggregateRoot> {
   ...
}
```

What do you think ? Does it makes sense ?

The main point are:
   - The state is concrete, the aggregate is a behaviour
   - any aggregate can be event sourced
   - repository can only manipulate aggregate roots as it is supposed to
   - aggregate can be aggregate root but not the other way around

Bernardstanislas: ok but I'm not a huge fan of the name `State`, it is confusing to me, what do you think about calling it directly with what it represents, `Contribution` in our case?

Bernardstanislas: Updated!

AnthonyBuisset (approved):
Looks very good !

PR #178, opened 2022-08-30
Event source application

Bernardstanislas:
This is done by introducing a business method on the `Contribution` aggregate, and using an `ApplicationProjector`.
@AnthonyBuisset I know you want to remove the mutability in the `apply_event` method, this can be done in another PR, this one is big enough :)


marketplace-core/src/application/contribution/apply.rs
```diff
@@ -38,6 +52,12 @@ impl Usecase for ApplyToContribution {
 		contribution_id: &ContributionId,
 		contributor_id: &ContributorId,
 	) -> Result<(), DomainError> {
-		self.contribution_service.apply(contribution_id, contributor_id)
+		let mut contribution = self.contribution_repository.find_by_id(contribution_id)?;
+		let events = contribution.apply(contributor_id)?;
+		self.event_store.append(&contribution.get_id(), events.to_owned())?;
```
tdelabro, line 57: is .to_owned() really necessary.
`events` is already a vec, and `append` also expects a vec

tdelabro: probably its just a `clone()` you need

Bernardstanislas: I wanted to clone it to use it later in the projectors, should I just use `.clone()`?

marketplace-domain/src/contribution/aggregate_root/mod.rs
```diff
@@ -13,14 +14,63 @@ pub use status::Status;
...
+			return Err(Error::AlreadyApplied(contributor_id.to_owned()));
+		}
+
+		let applied_event = Event::Applied {
+			id: self.id.to_owned(),
+			contributor_id: contributor_id.to_owned(),
+		};
+
+		self.apply_event(&applied_event);
+		Ok(vec![applied_event])
```
tdelabro, line 53: Weren't we supposed to let the caller the responsibility to apply or not the events on the instance he is manipulating. At the same time he chose to store or note those events in the store/bus.

Bernardstanislas: Why wouldn't we want to apply them in the method directly? I know @AnthonyBuisset had the same remark so maybe we can talk about this IRL.
And for the second remark, the caller MUST ALWAYS store the events and pass them to the bus.

marketplace-domain/src/contribution/aggregate_root/mod.rs
```diff
@@ -13,14 +14,63 @@ pub use status::Status;
...
+	}
+
+	pub fn get_id(&self) -> Id {
+		self.id.to_owned()
+	}
+
+	pub fn get_status(&self) -> ContributionStatus {
+		self.status.to_owned()
+	}
```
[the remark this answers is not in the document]
Bernardstanislas: Oh sure, I'm changing it

marketplace-domain/src/contribution/aggregate_root/mod.rs
```diff
@@ -41,6 +91,12 @@ impl Aggregate for Contribution {
 				self.gate = gate;
 				self.status = Status::Open;
 			},
+			Event::Applied {
+				id: _,
+				contributor_id,
+			} => {
+				self.applicants.push(contributor_id.to_owned());
```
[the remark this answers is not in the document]
Bernardstanislas: Ok you're right, I'm changing it

marketplace-indexer/src/domain/obervers/contribution_event_store_logger.rs
```diff
@@ -17,6 +17,10 @@ impl<ES: EventStore<Contribution>> Observer for ES {
 				id,
 				contributor_id: _,
 			} => id,
+			ContributionEvent::Applied {
```
tdelabro, line 20: Do we still need this logger?
We are now storing them in db

marketplace-domain/src/contribution/aggregate_root/mod.rs
tdelabro: Well, nothing enforces this.
If he drops its aggregate and don't pass the event, it's like nothing happened.

marketplace-domain/src/contribution/projections/application.rs
```diff
@@ -31,15 +31,19 @@ impl Display for Status {
...
 }
 
-impl Application {
-	pub fn new(
+impl Projection {
+	pub fn new(id: Id, contribution_id: ContributionId, contributor_id: ContributorId) -> Self {
+		Self::new_with_status(id, contribution_id, contributor_id, Status::Pending)
+	}
+
+	pub fn new_with_status(
```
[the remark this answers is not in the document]
Bernardstanislas: We sure can but it is used in the db implementation to convert db models to the projection struct. What is the best practice in this case?

marketplace-indexer/src/domain/obervers/contribution_event_store_logger.rs
Bernardstanislas: This is still useful in the early times of our switch to event sourcing, it will help us debug and monitor but I agree this will need to be removed later

marketplace-domain/src/contribution/aggregate_root/mod.rs
Bernardstanislas: yes and that disturbs me :)

marketplace-indexer/src/domain/obervers/contribution_event_store_logger.rs
tdelabro: ok no problem

marketplace-domain/src/contribution/projections/application.rs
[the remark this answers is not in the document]
tdelabro: I would use the `From` trait, a keep this possibility concealed inside the trait.
If we only store Projections that are built from events we know they are valid, so we can confidently transform them back into arbitrary Projections.

This way we can remove it from the trait itself. And close this risky door :)

marketplace-domain/src/contribution/aggregate_root/mod.rs
tdelabro: We can imagine some use-cases where it could be useful to know what event would be generated beforehand.
For dry runs, or something like this.

marketplace-domain/src/aggregate.rs
```diff
@@ -1,8 +1,14 @@
...
 	type Event;
 	fn apply_event(&mut self, event: &Self::Event);
-	fn from_events(events: Vec<Self::Event>) -> Self;
+	fn from_events(events: Vec<Self::Event>) -> Self {
+		let mut contribution = Self::default();
+		events.iter().for_each(|event| {
+			contribution.apply_event(event);
+		});
+		contribution
+	}
```
ofux, line 11: This feel odd to have a default implem for contributions in the Aggregate trait.

ofux (approved):

marketplace-domain/src/aggregate.rs
Bernardstanislas: Yes but I could not think of another way to have an empty shell to start the event sourcing from

marketplace-core/src/routes/applications/list/tests.rs
```diff
@@ -34,6 +34,14 @@ impl ApplicationProjectionRepository for EmptyDatabase {
 		unimplemented!()
 	}
 
+	fn find_by_contribution_and_contributor(
```
AnthonyBuisset, line 37: shouldn't we be using a mock instead of adding all those unimplemented methods ?

marketplace-domain/src/contribution/projectors/application.rs
```diff
@@ -0,0 +1,161 @@
...
+				self.application_projection_repository.create(application)
+			},
+		}
+	}
+}
+
+impl Projector<Contribution> for ApplicationProjector {
+	fn project(&self, event: &<Contribution as Aggregate>::Event) {
+		let result = match event {
+			ContributionEvent::Applied {
```
AnthonyBuisset, line 84: this could be turned into an `if let`

marketplace-domain/src/contribution/projectors/application.rs
```diff
@@ -41,6 +41,28 @@ impl ApplicationProjector {
...
+		contributor_id: &ContributorId,
+	) -> Result<(), ApplicationProjectionRepositoryError> {
+		let mut contribution_applications = self
+			.application_projection_repository
+			.list_by_contribution(contribution_id, None)?;
+		contribution_applications
+			.iter_mut()
+			.map(|application| {
+				match application.contributor_id() {
+					id if id == contributor_id =>
```
AnthonyBuisset: is this logic a domain rule or just for the projection ?

AnthonyBuisset (approved):
Great work !
Let a few comments to think about

marketplace-domain/src/contribution/projectors/application.rs
```diff
@@ -63,6 +63,23 @@ impl ApplicationProjector {
...
+		contribution_id: &ContributionId,
+	) -> Result<(), ApplicationProjectionRepositoryError> {
+		let mut contribution_applications = self
+			.application_projection_repository
+			.list_by_contribution(contribution_id, None)?;
+		contribution_applications
+			.iter_mut()
+			.map(|application| {
+				application.set_status(ApplicationStatus::Pending);
+				self.application_projection_repository.update(application.to_owned())
```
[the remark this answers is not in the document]
Bernardstanislas: I did not want to put them as pending since they were rejected, so it would be weird for a contributor to be notified that they were rejected then pending again. I'm gonna ask to the product team :)

PR #179, opened 2022-08-30
deduplicate events

ofux:
- ⬆️ upgrade apibara proto to same version as apibara server
- ✨ introduce StorableEvent to allow event deduplication
- ♻️ use ObservedEvent::default in tests
- 🚨 clippy
- ✨ deduplicate events in event_store


marketplace-infrastructure/src/database/domain_implementations/event_store.rs
```diff
@@ -25,9 +38,13 @@ impl EventStore<Contribution> for Client {
 			})
 			.collect::<Result<Vec<_>, EventStoreError>>()?;
 
-		diesel::insert_into(events::table)
-			.values(&events)
-			.execute(&*connection)
+		connection
+			.transaction(|| {
```
Bernardstanislas, line 43: ❤️

Bernardstanislas (approved):
Amazing! You should configure your editor formatter though 😆

Bernardstanislas (commented):
After a second thought, I think it would be cleaner if we keep the notion of deduplication to the indexer perimeter.
I think the event_store shouldn't know about deduplication, only the indexer should.
The indexer would open a transaction, store the deduplicatons with its own ways and store the events with the event_store.

marketplace-indexer/src/infrastructure/apibara/events/mod.rs
```diff
@@ -59,6 +69,11 @@ mod test {
 	use marketplace_domain::ContributionEvent;
 	use rstest::*;
 
+	const CONTRACT_ADDRESS: &str = "0x42";
```
AnthonyBuisset, line 72: using const here is a great idea to avoid too many fixtures ❤️

marketplace-indexer/src/infrastructure/apibara/events/mod.rs
```diff
@@ -45,7 +48,14 @@ impl TryFrom<ApibaraEvent> for DomainEvent {
 					_ if selector == contribution::Validated::selector() =>
 						Ok(contribution::Validated::to_domain_event(data)?),
 					_ => Err(Self::Error::Unsupported),
-				}
+				}?;
+
+				let address = std::str::from_utf8(&address).unwrap();
```
AnthonyBuisset, line 53: You can use ContractAddress type as well

AnthonyBuisset (approved):
As discussed, to be refactored/re-designed later to remove deduplication ids from the domain

Bernardstanislas (approved):

AnthonyBuisset (approved):
LGTM!

PR #184, opened 2022-08-31
feat: update contribution contract 'new_contribution' call signature

tdelabro:


marketplace-core/src/bin/e2e/contribution/add.rs
```diff
@@ -1,7 +1,7 @@
 use super::*;
 use marketplace_core::dto::ContributionCreation;
 
-pub async fn add_contribution(github_issue_number: i64, project_id: u64, gate: u8) {
+pub async fn add_contribution(github_issue_number: u64, project_id: u64, gate: u8) {
 	println!("Adding contribution {project_id}/{github_issue_number} with gate {gate}");
```
tdelabro, line 5: Should we keep this `println!`? Shouldn't we use `debug!` for this?

Bernardstanislas: Sure, but this is not very important since it's the e2e binary we are talking about here

tdelabro: ok

AnthonyBuisset (approved):

PR #185, opened 2022-08-31
Improve generic code

tdelabro:


marketplace-indexer/src/infrastructure/apibara/indexing_service.rs
```diff
@@ -14,11 +13,11 @@ use super::{
 use crate::domain::*;
 
 #[async_trait]
-impl IndexingService for Client {
+impl IndexingService<BlockchainObserverComposite> for Client {
```
AnthonyBuisset, line 16: the purpose of the composite design pattern is to not know that it is a composite and use it as a regular observer

AnthonyBuisset: SO maybe implementing for `O: Observer + Clone` ? but cloning an observer is not the same as cloning an Arc<> which just clone the pointer and is more efficient IMO

marketplace-indexer/src/main.rs
```diff
@@ -83,14 +83,13 @@ fn build_contribution_observers(
...
-		database.confirmed(confirmation_blocks_count),
+		database.clone().confirmed(confirmation_blocks_count),
 		Arc::new(ContributionObserver::new(Arc::new(contribution_projector)))
 			.confirmed(confirmation_blocks_count),
 		Arc::new(ApplicationObserver::new(Arc::new(contribution_service)))
 			.confirmed(confirmation_blocks_count),
-	]);
-
-	Arc::new(observer)
+		database.confirmed(confirmation_blocks_count),
```
AnthonyBuisset, line 93: why adding it twice ?

PR #186, opened 2022-09-01
add timeout for github requests and remove confirmation observer

AnthonyBuisset:
- :bug: add timeout management in github client call to avoid blocking request
- :mute: remove debugging logs
- :bug: increase log channel size in local to avoid losing logs
- :recycle: remove confirmation observer as not needed on StarkNet
- :hammer: rename docker containers and database in local
- :hammer: remove starknet-devnet from docker-compose as not used for now
- :recycle: move event store observer first in the composite


.env.example
```diff
@@ -7,6 +7,7 @@ APIBARA_URL="http://localhost:7171"
 export API_KEY=ROOT
 export API_URL="http://localhost:8000"
 export LOGS=terminal
+SLOG_CHANNEL_SIZE=1024
```
tdelabro, line 10: add export before the line

tdelabro (approved):

.env.example
AnthonyBuisset: Done

PR #191, opened 2022-09-02
Migrate pending applications to events

Bernardstanislas:


migrations/2022-09-02-140917_event-source-applications/down.sql
```diff
@@ -0,0 +1,5 @@
+UPDATE applications SET status = 'pending'
```
AnthonyBuisset, line 1: should't the `down.sql` be a `delete` to be an exact revert of the `up.sql` ?

Bernardstanislas: oh sure

Bernardstanislas: done

AnthonyBuisset (approved):

PR #192, opened 2022-09-02
refactor: aggregate root repository logic direct implem instead of trait

tdelabro:


marketplace-domain/src/services/contribution.rs
```diff
@@ -151,7 +161,7 @@ mod test {
 
 	#[rstest]
 	fn application_success(
-		mut contribution_repository: MockRepository<Contribution>,
+		mut contribution_repository: Repository<Contribution>,
```
AnthonyBuisset, line 164: I think we can actually use MockRepository here and it should still work based on `mockall_double` implem ?

tdelabro: Yes indeed. It work

marketplace-domain/src/contribution/projectors/contribution/contribution_with_github_data_projector.rs
```diff
@@ -112,7 +112,7 @@ impl Projector<Contribution> for WithGithubDataProjector {
 				__self.on_assign(id, contributor_id),
 			ContributionEvent::Unassigned { id } => __self.on_unassign(id),
 			ContributionEvent::Validated { id } => __self.on_validate(id),
-			ContributionEvent::Applied { id, contributor_id } => todo!(),
+			ContributionEvent::Applied { .. } => todo!(),
```
AnthonyBuisset, line 115: I did not know this, nice !

AnthonyBuisset (approved):

AnthonyBuisset:
@ofux @Bernardstanislas You should check this, it is a nice way of implementing our domain logic

Bernardstanislas (commented):
So if I understand correctly, we no longer rely on traits as dependencies, but on concrete impl?
Second question, what about sharing the services instances between threads with `Arc`s? I see we clone for `AggregateRootRepository` but not for other services, do we want to introduce this new way of injection?

PR #193, opened 2022-09-05
refacotring proposal for aggregates management

AnthonyBuisset:
* Differentiate Aggregate State from Aggregate behavior
* Make very clear that a Contribution is both a State and an Aggregate (see `type State = Self;`)
* Pass state as value in `apply()` so it cannot be used anymore and forces to be event sourced again

- :recycle: make Aggregate state immutable
- :recycle: pass events as reference in aggregate
- :recycle: remove event application from aggregate command as per SRP


marketplace-domain/src/contribution/aggregate_root/mod.rs
```diff
@@ -76,40 +75,46 @@ impl Contribution {
 impl Aggregate for Contribution {
 	type Event = Event;
 	type Id = Id;
+	type State = Self;
```
Bernardstanislas, line 78: I think we will do this for all aggregates, so do we even need this `State` type?
`apply_event` could take an immutable `&self` and return a new instance of `Self`, right? This would bring the immutability you are looking for, without the overhead of `State`

AnthonyBuisset: yes we can, but I like how clear it is that we make the assumption of
```
type State = Self;
```
otherwise I feel we are mixing stuff

tdelabro: I don't feel like we need this.
Self is already present and really expressive.
No need to complicate things with a ditto type

marketplace-domain/src/contribution/aggregate_root/mod.rs
```diff
@@ -128,6 +133,11 @@ mod test {
 		Id::from_str("0xfa").unwrap()
 	}
 
+	#[fixture]
```
tdelabro, line 136: Tests in a separated file

tdelabro (changes requested):

marketplace-core/src/application/contribution/apply.rs
```diff
@@ -60,7 +60,8 @@ impl Usecase for ApplyToContribution {
 		contribution_id: &ContributionId,
 		contributor_id: &ContributorId,
 	) -> Result<(), DomainError> {
-		let mut contribution = self.contribution_repository.find_by_id(contribution_id)?;
+		let contribution = self.contribution_repository.find_by_id(contribution_id)?;
+		let contribution_id = contribution_id.to_owned();
```
tdelabro, line 64: Why do you have to own it?
You are using it as a reference on l73.

marketplace-domain/src/aggregate.rs
```diff
@@ -1,13 +1,15 @@
...
-		});
-		contribution
+
+	fn apply_event(self, event: &Self::Event) -> Self;
+
+	fn apply_events(self, events: &Vec<Self::Event>) -> Self {
+		events.iter().fold(self, Self::apply_event)
+	}
+
+	fn from_events(events: &Vec<Self::Event>) -> Self {
```
tdelabro, line 11: instead of `&Vec<Self::Event>` use `&[Self::Event]` for more versatile/generic code

marketplace-domain/src/contribution/aggregate_root/mod.rs
```diff
@@ -36,9 +36,9 @@ pub struct Contribution {
...
 
 impl Contribution {
-	pub fn apply(&mut self, contributor_id: &ContributorId) -> Result<Vec<Event>, Error> {
+	pub fn apply(self, contributor_id: &ContributorId) -> Result<Vec<Event>, Error> {
 		if self.status != Status::Open {
-			return Err(Error::CannotApply(self.status.to_owned()));
+			return Err(Error::CannotApply(self.status));
 		}
 		if self.applicants.contains(&contributor_id) {
 			return Err(Error::AlreadyApplied(contributor_id.to_owned()));
```
tdelabro, line 44: use `.clone()` when you want to clone to the same type.
`.to_owned()` is susceptible to change return a type which is not the one of you original value.

`to_owned` does more than `clone`. So if you just need to `clone` use `clone`

marketplace-domain/src/contribution/aggregate_root/mod.rs
```diff
@@ -77,39 +76,44 @@ impl Aggregate for Contribution {
...
+				status: Status::Open,
+				..self
 			},
 			Event::Applied {
 				id: _,
 				contributor_id,
-			} => {
-				self.applicants.push(contributor_id.to_owned());
+			} => Self {
+				applicants: [&self.applicants[..], &[contributor_id.to_owned()]].concat(),
```
tdelabro, line 98: You are allocating a new vector there.
I think there is a way to move the original vector ownership from `self` to the new `Self` and then push the new applicant as it was done before.

This spare some allocation

tdelabro: It should be doable in two lines

marketplace-domain/src/contribution/aggregate_root/mod.rs
AnthonyBuisset: if I have a reference (&) and I want to dereference (having the owned type) I have to call `to_owned` right ?
for instance in this particular case it is ok as the event owns the contributor_id which is passed by reference in the function.
Am I right ?

tdelabro (approved):

PR #205, opened 2022-09-07
React to contribution claimed

tdelabro:


marketplace-indexer/src/infrastructure/apibara/events/contribution/claimed.rs
```diff
@@ -0,0 +1,78 @@
...
+pub struct Claimed;
+
+impl EventTranslator for Claimed {
+	fn selector() -> FieldElement {
+		get_selector_from_name("ContributionClaimed").unwrap()
+	}
+
+	fn to_domain_event(mut topics: Topics) -> Result<Event, FromEventError> {
+		let contribution_id: HexPrefixedString = topics.pop_front_as()?;
+		let contributor_id: ContributorId = topics.pop_front_as()?;
```
Bernardstanislas, line 18: I'm really sorry because we discussed this afternoon with @ofux and @AnthonyBuisset about this, but we prefer not to talk about contributor ids from now on, but rather contributor accounts that will be `HexPrefixedString`s.
The contribution smart contract does not know about contributor ids, and shouldn't, so it will emit events with the contributor account directly. And we'll integrate in a next iteration the removal of contributor ids

tdelabro: Ok. So in the meantime, it will be the only one using account and not contributor.
Works for me, no worries.

marketplace-indexer/src/infrastructure/apibara/indexing_service/mod.rs
```diff
@@ -0,0 +1,148 @@
+#[cfg(test)]
```
AnthonyBuisset, line 1: Could we agree to follow this kind of pattern : https://github.com/onlydustxyz/marketplace-backend/pull/204/files#diff-96e13d4d42515c5d7a618a623eb9e644ea055bd921cc72a64c6c2296e2b5e627 ?
This is for me a good compromise:
* Tests in a dedicated file
* filename is still relevant (more than `mod.rs`)

marketplace-indexer/src/infrastructure/apibara/events/contribution/claimed.rs
AnthonyBuisset: ProjectMember events are also using accounts instead of ids

marketplace-domain/src/contribution/projectors/project/mod.rs
```diff
@@ -1,5 +1,64 @@
-mod project;
-pub use project::*;
-
 #[cfg(test)]
-mod test;
+mod tests;
```
AnthonyBuisset, line 2: I prefer the old way (with the clippy ignore) as otherwise all files will be names either `mod.rs` or `test.rs` and this is way more disturbing to me than having tests in the same file as the code

marketplace-indexer/src/infrastructure/apibara/events/contribution/claimed.rs
Bernardstanislas: Yes, progressively removing the references to ids

marketplace-indexer/src/infrastructure/apibara/indexing_service/mod.rs
tdelabro: filename cannot be the same as the directory one. It's discouraged by clippy

marketplace-domain/src/contribution/projectors/project/mod.rs
tdelabro: all files are already named contributions, you cannot find the one you are looking for just by its name. You always have to look for the whole path. 
So it really doesn't bother me
Screenshot 2022-09-08 at 11 04 48

marketplace-core/src/routes/applications/list/tests.rs
```diff
@@ -108,11 +108,7 @@ impl ApplicationProjectionRepository for FilledDatabase {
 						0.into(),
 						contributor_id.clone(),
 					),
-					ApplicationProjection::new(
-						Uuid::from_u128(3).into(),
-						1.into(),
-						contributor_id.clone(),
-					),
+					ApplicationProjection::new(Uuid::from_u128(3).into(), 1.into(), contributor_id),
```
AnthonyBuisset, line 111: I tend to remove this kind of new as I do not think it brings value. It also allows to make mistakes like:
```
ApplicationProjection::new(id, contribution_id, contributor_id)
```
vs
```
ApplicationProjection::new(id, contributor_id, contribution_id)
```

we are not always protected by the type system.

Whereas with this syntax, it is safer and easier to detect mistakes like:
```
ApplicationProjection {
   id, 
   contribution_id: contributor_id,
   contributor_id: contribution_id,
}
```

tdelabro: It implies that all fields of the struct are public. And can be set to arbitrary values within the code.
It's ok to let this possibility for some usecases, not for others.
But when we can, you are right, this is  more expressive

PR #206, opened 2022-09-08
fetch project data from github at event projection time instead of GET

AnthonyBuisset:
- :recycle: move project list route in a dedicated file
- ✨ fetch additional info from github and store in projects table
- ⚰️ remove unused github model
- ♻️ fetch data from projection in GET
- ⚰️ remove unused RepoCache


ofux (approved):
LGTM

PR #207, opened 2022-09-08
feature/list members

ofux:
- ✨ return member list in /projects
- ✅ it's better when tests actually test something :)


Bernardstanislas (approved):

AnthonyBuisset:
LGTM

Instead of adding an infrastructure test, I would suggest to complete the component test I wrote earlier and could not finish dur to this missing feature:
https://github.com/onlydustxyz/marketplace-backend/blob/main/marketplace-core/src/application/project/test.rs#L79

AnthonyBuisset (approved):

PR #208, opened 2022-09-08
add refresh instance on main branche

SamuelKer:
add refresh instance


AnthonyBuisset (commented):
Won't the new job run in parallel with the other 2 ?
I think we should add a kind of trigger to run it only once the docker images have been uploaded

AnthonyBuisset (approved):

PR #212, opened 2022-09-09
feat: add past lead contributors indexer

tdelabro:


AnthonyBuisset (approved):

marketplace-indexer/src/domain/observers/contribution.rs
```diff
@@ -1,18 +1,30 @@
+use std::marker::PhantomData;
+
 use super::*;
-use std::sync::Arc;
 
-pub struct ContributionObserver<P: Projection<A = Contribution>> {
-	projector: Arc<dyn Projector<P>>,
+pub struct ContributionObserver<
+	PJTR: Projector<PJTN>,
```
Bernardstanislas, line 6: This disturbs me a lot, it's a kind of uppercased shorten name, it implies it is an acronym but it's not.
I understand you did this to avoid the collision with the `Projector` trait, what about `ProjectorType` instead?

marketplace-indexer/src/domain/observers/contribution.rs
```diff
@@ -1,18 +1,30 @@
...
+	PJTN: Projection<A = Contribution> + Send + Sync,
+> {
+	projector: PJTR,
+	projection_type: PhantomData<PJTN>,
```
Bernardstanislas, line 10: Why is this one suffixed with `_type` where the other is not?

tdelabro: this one is a phantom field. It is not sized and does not represent any actual data. It exists just to give a type information to the compiler, but totally disappears in the compiled program. 

So this field just holds the information about the type of the projection used in the struct (but not an actual instance of this type).
On the other hand, the field `projector` hold a projector struct.

marketplace-indexer/src/domain/observers/contribution.rs
tdelabro: Sure. Rather, `Projector` and `Projection`

tdelabro: Outch, it collides with the trait names.

tdelabro: `Pjtr`/ `Pjtn` ?
 `ProjectorT` / `ProjectionT` ?

PR #216, opened 2022-09-12
fix: applications status are updated in a single transaction

tdelabro:


marketplace-domain/src/repositories/application_projection.rs
```diff
@@ -34,4 +34,16 @@ pub trait Repository: Send + Sync {
 		&self,
 		contributor_id: Option<ContributorId>,
 	) -> Result<Vec<ApplicationProjection>, Error>;
+	fn for_a_contribution_set_all_status(
```
Bernardstanislas, line 37: I am not in favor of putting business-intensive methods in repositories

AnthonyBuisset:
The code coverage is decreasing. Ca you please check ? It may due to dead code appearing or missing tests

PR #217, opened 2022-09-13
refactor: Projector trait became a DomainEventHandler

tdelabro:


marketplace-indexer/src/domain/observers/contribution.rs
```diff
@@ -3,14 +3,14 @@ use std::marker::PhantomData;
 use super::*;
 
 pub struct ContributionObserver<
-	TProjector: Projector<TProjection>,
+	TProjector: EventHandler<TProjection>,
```
Bernardstanislas, line 6: Do we keep the term projector then?

Bernardstanislas (approved):

marketplace-indexer/src/domain/observers/contribution.rs
tdelabro: changed

ofux (approved):

marketplace-domain/src/contribution/aggregate_root/mod.rs
```diff
@@ -62,55 +61,58 @@ impl Contribution {
...
-			},
-			Event::Unassigned { id: _ } => Self {
-				status: Status::Open,
-				contributor_id: None,
-				..self
-			},
-			Event::Validated { id: _ } => Self {
-				status: Status::Completed,
-				..self
+			Event::Contribution(contribution_event) => match contribution_event {
```
AnthonyBuisset, line 71: @tdelabro This update feels weird to me.
We are losing safety here, an aggregate can be even sourced from any domain event ?

PR #223, opened 2022-09-14
decorelate indexer creations from events listeners to prepare for dynamic indexing

AnthonyBuisset:
- ♻️ refactor error management in indexer
- ✨ add indexer listing in repository
- ♻️ refactor indexer main file
- ♻️ fetch events from all indexers in parallel

### DIFF
--- a/marketplace-indexer/src/application/builders/indexer.rs
+++ b/marketplace-indexer/src/application/builders/indexer.rs
@@ -101,6 +101,7 @@ impl Builder {
 #[cfg(test)]
 mod tests {
 	use super::*;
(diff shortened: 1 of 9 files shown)


Bernardstanislas (approved):

ofux (approved):
LGTM, but shouldn't there be a few new tests? Especially about the `index_events` function, which btw would probably better be outside of `main.rs`?

AnthonyBuisset:
> LGTM, but shouldn't there be a few new tests? Especially about the `index_events` function, which btw would probably better be outside of `main.rs`?

Indeed, will do

PR #224, opened 2022-09-15
Webhook

tdelabro:


marketplace-infrastructure/src/event_webhook/mod.rs
```diff
@@ -21,12 +30,17 @@ impl EventListener for EventWebHook {
 				},
 			};
 
-			let client = reqwest::Client::new();
-			if let Err(e) = client.post(target.clone()).json(&event).send().await {
-				error!("Failed to send event to hook target '{target}': {e}");
+			match self.0.post(target.clone()).json(&event).send().await {
```
Bernardstanislas, line 33: Can we put an explicit attribute instead of directly wrapping the `Client`? `self.0` is not clear, `self.client` would be better

Bernardstanislas (changes requested):

marketplace-infrastructure/src/event_webhook/mod.rs
tdelabro: done

marketplace-infrastructure/src/event_webhook/mod.rs
```diff
@@ -0,0 +1,48 @@
...
+
+impl EventWebHook {
+	pub fn new(client: reqwest::Client) -> Self {
+		Self { web_client: client }
+	}
+}
+
+#[async_trait]
+impl EventListener for EventWebHook {
+	async fn on_event(&self, event: &Event) {
```
AnthonyBuisset, line 23: I would rather refactor this into a faillible function so that we can leverage the `?` syntax.
Then log and ignore the error in the `on_event`
This would limit the nesting

marketplace-infrastructure/src/event_webhook/tests.rs
```diff
@@ -0,0 +1,92 @@
+use std::ffi::OsString;
```
AnthonyBuisset, line 1: I thought we agreed to have tests either:
* in the same file for unit tests (leveraging the fixture to limit the size)
* in a dedicated test file with proper naming (e.g. component test) at the root of a crate
This one looks like a unit test to me as dependencies are mocked

marketplace-infrastructure/src/event_webhook/tests.rs
```diff
@@ -0,0 +1,92 @@
...
+		assert_eq!(captured_logs[0].level, Level::Info);
+	});
+}
+
+#[tokio::test]
+#[serial]
+async fn env_variable_invalid() {
+	testing_logger::setup();
+	let _lock = lock_test();
+	let _test = set_env(OsString::from(WEBHOOK_TARGET_ENV_VAR), "Some random junk");
```
AnthonyBuisset, line 41: nice way to silo the test impacts, we should leverage that as a macro or fixture in the future

marketplace-infrastructure/src/event_webhook/tests.rs
```diff
@@ -0,0 +1,92 @@
...
+	testing_logger::setup();
+	let _lock = lock_read();
+	let webhook = EventWebHook(reqwest::Client::new());
+
+	let event = Event::Contribution(ContributionEvent::Validated {
+		id: Default::default(),
+	});
+	webhook.on_event(&event).await;
+
+	testing_logger::validate(|captured_logs| {
```
AnthonyBuisset, line 25: I think that asserting the call is not made and test in not panicing is more important than the log printed.
It would allow to refactor into a same test several inputs using the `#[case]` pattern and reduce the test size

AnthonyBuisset (changes requested):

marketplace-infrastructure/src/event_webhook/mod.rs
tdelabro: Not sure about that. 
Listeners are by definition infallible.
You put stuff in it, maybe things happen, maybe not. You don't care and continue your main execution.

marketplace-infrastructure/src/event_webhook/tests.rs
tdelabro: 93 lines of test for vs 48 of actual code.
It started to feel too long for me

marketplace-infrastructure/src/event_webhook/tests.rs
tdelabro: Not sure we want to abstract it more.
The`testing_logger` lib offers a variety of functionalities, we won't always be using this exact combination.

If your really want to do so, do it into the lib itself.

marketplace-infrastructure/src/event_webhook/tests.rs
tdelabro: I cannot test that reqwest did not do the call.
What I can do is, like in the very last test, mock a webserver and count the number of times the route was called.
But if the env variable is not set, there is, by definition, no route to monitor.
So I still cannot know that the call was not made.

Bernardstanislas (approved):
Ok for my remark, I let you continue with @AnthonyBuisset 's returns

marketplace-infrastructure/src/event_webhook/tests.rs
tdelabro: and it protects you when you do global searches, so you don't have to look through tons of tests when trying to navigate the project

Screenshot 2022-09-15 at 12 16 10

Screenshot 2022-09-15 at 12 16 23

marketplace-infrastructure/src/event_webhook/tests.rs
AnthonyBuisset: ok

marketplace-infrastructure/src/event_webhook/tests.rs
AnthonyBuisset: then checking the test does not panic is enough for me.
assertions on logs is a bit too much IMO.

marketplace-infrastructure/src/event_webhook/tests.rs
AnthonyBuisset: We should all agree on a standard and stick to it, otherwise the repo won't be consistent and subjective.
But we do not have such standard yet, so I am approving the PR as is until we do

.github/workflows/tests.yml
```diff
@@ -26,18 +26,18 @@ jobs:
...
-        with:
-          path: |
-            ~/.cargo/bin/
-            ~/.cargo/registry/index/
-            ~/.cargo/registry/cache/
-            ~/.cargo/git/db/
-            target/
-          key: ${{ runner.os }}-cargo-${{ hashFiles('**/Cargo.lock') }}
-          restore-keys: ${{ runner.os }}-cargo-
+      # - name: Set up cargo cache
```
AnthonyBuisset, line 29: why is it not needed anymore ? to be uncommented or deleted ?

tdelabro: I'm trying stuff because CI is failing. But it's not working either

marketplace-infrastructure/src/event_webhook/mod.rs
tdelabro: done

marketplace-infrastructure/src/event_webhook/tests.rs
tdelabro: solved with factorisation in another function

PR #231, opened 2022-09-16
Add applied_at in Applied events

ofux:
- ✨ add applied_at in applications
- ✅ 🐛 fix component test failling due to too many connections to the DB
- 🗃️ migrate Applied events

### DIFF
--- a/marketplace-core/src/application/contribution/refresh/test.rs
+++ b/marketplace-core/src/application/contribution/refresh/test.rs
@@ -1,6 +1,6 @@
 use super::*;
 use async_trait::async_trait;
-use chrono::Utc;
+use chrono::{NaiveDate, NaiveDateTime, Utc};
 use dotenv::dotenv;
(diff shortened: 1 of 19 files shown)


marketplace-core/src/application/contribution/refresh/test.rs
```diff
@@ -53,11 +53,17 @@ impl Storable for ContributionEvent {
 }
 
 #[fixture]
-fn database() -> Arc<DatabaseClient> {
+#[once]
```
Bernardstanislas, line 56: 🤯

Bernardstanislas (approved):

marketplace-core/src/application/contribution/refresh/test.rs
ofux: I just rebased and saw @tdelabro basically did the same at the same time. Les grands esprits se rencontrent :)

PR #239, opened 2022-09-21
Fix project projection bug

tdelabro:
Here I revert the try to have generic CRUD methods. It is impossible to upsert, and require a lot of `where` clause for insert and update. And just make it slightly more readable.
The truth it that diesel is built in a way that discourages this by design. The current state of the code is how it is supposed to be used. If we want to spare some redundant code, we will have to generate code through macro. Totally doable, and a good use case for somebody who when to experiment with it.

The rest of the PR is a mix of:
- renaming every projection repo to the `insert`, `update`, `upsert` naming convention
- get rid of unnecessary checks of existence when receiving events that by design require the aggregate to exist
- get rid of repositories methods that are never used
- get rid of test that tests nothing but the fact that the code actually calls the methods he is calling (testing the entity implementation details rather that the entity logical behavior)

I tested to refresh the project projection locally with the dump DB. It worked. So I guess the bug that was present in staging is solved.


marketplace-domain/src/contribution/projectors/application.rs
```diff
@@ -26,70 +26,38 @@ impl ApplicationProjector {
...
-					self.uuid_generator.new_uuid().into(),
-					contribution_id.clone(),
-					contributor_id.clone(),
-					Utc::now().naive_utc(),
-				)
-				.into_refused();
-				self.application_projection_repository.create(application)
-			},
-		}
+		self.application_projection_repository.update_status(
```
AnthonyBuisset, line 44: again, some changes as this will update ALL applications to refused, should not have side effect right now as there should be only one pending application at a time, but still...
Maybe we should consider using the application id in the event/aggregate ?

marketplace-domain/src/contribution/projectors/application.rs
```diff
@@ -26,70 +26,38 @@ impl ApplicationProjector {
...
-					self.uuid_generator.new_uuid().into(),
-					contribution_id.clone(),
-					contributor_id.clone(),
-					Utc::now().naive_utc(),
-				)
-				.into_accepted();
-				self.application_projection_repository.create(application)
-			},
-		}
+		self.application_projection_repository.update_status(
```
AnthonyBuisset, line 56: same comment here, all application (even the previously rejected ones) gets accepted

marketplace-domain/src/contribution/projectors/application.rs
```diff
@@ -96,284 +96,3 @@ impl EventListener for ApplicationProjector {
 		}
 	}
 }
-
```
AnthonyBuisset, line 131: that are all the tests that got broken with the change

marketplace-core/src/routes/projects/mod.rs
```diff
@@ -35,7 +35,7 @@ pub async fn new_project(
 		url: project.html_url,
 		logo_url: Some(owner.avatar_url),
 	};
-	ProjectProjectionRepository::store(database.as_ref(), projection)
+	ProjectProjectionRepository::insert(database.as_ref(), projection)
```
AnthonyBuisset, line 38: the bug in staging is here => we use to upsert a project in this function, now we insert and fail if already exists. But this function will soon disappear I guess, so no big deal for me.
Maybe there was never a real bug in staging

AnthonyBuisset (changes requested):

AnthonyBuisset:
> The current state of the code is how it is supposed to be used.

DO you have some links to explain what is the expected way of using diesel ?
because I have seen several ways to do the same thing: where to put the `.filter()`, `diesel::update(table)` vs `table.update()`...

AnthonyBuisset:
> we will have to generate code through macro

That is an excellent idea IMO. If we base our design of DB tables on a standard (like must have an ID, etc) we could generate all boilerplate code to perform simple CRUD operations on all our repositories.
Downside is that we would generate useless code sometimes, but as it is generated, I guess this is worth it

marketplace-domain/src/services/contributor.rs
```diff
@@ -12,7 +11,7 @@ pub enum Error {
 	Infrastructure(anyhow::Error),
 }
 
-#[cfg_attr(test, automock)]
+#[automock]
```
AnthonyBuisset, line 14: I would rather move the opposite way and have mockall dependency only in `cfg(test)`

migrations/2022-09-22-084844_simplify-application-projection-table/up.sql
```diff
@@ -0,0 +1,8 @@
+DELETE FROM applications
+    WHERE "status" != 'pending';
+
+ALTER TABLE applications
+    DROP COLUMN "status";
```
AnthonyBuisset, line 5: is it possible to remove the ID column as well ?

marketplace-domain/src/services/contributor.rs
tdelabro: Yes but it just doesn't work.
If the trait is reexported under another name, the compiler lose track of it and cannot import it anymore.

migrations/2022-09-22-084844_simplify-application-projection-table/up.sql
tdelabro: yup. I'm on it

AnthonyBuisset (approved):

PR #241, opened 2022-09-22
🐛 reduce gRPC keepalive to avoid closing the stream

AnthonyBuisset:
When not enough messages are sent in the gRPC stream, the stream might be closed by the protocol.
In that case, the back-end panics and gets restarted by Heroku.

This PR reduces the keepalive interval to keep the stream opened.

However, there is a bug currently in `apibara` node:
> when asking for a block in the future (can happen we client restarts and there is no block since last stream, or if node is not synced with the gateway),
> the node sends nothing in the stream and hangs indefinitely, until we restart the connection.

So I have mixed feelings about this fix:
1. either wait for apibara fix to load it and rely on heroku restart in case of issue during a SW load
2. load the fix and monitor SW loads on heroku (we should enforce a downtime of the indexer of at least one block or restart manually)


AnthonyBuisset:
Not needed

PR #252, opened 2022-09-26
Store events metadata

tdelabro:


Bernardstanislas (commented):
Two remarks:
- what is the goal of the `origin` field? Can't it be included in the `metadata` json?
- I think the concept of event metadata should only exist at the infrastructure layer. It shouldn't live in the application or domain layers, it is not business related, it only acts as a trace for audit sake

tdelabro:
`what is the goal of the origin field? Can't it be included in the metadata json?`

The idea behind the origin field is to discriminate a bit between the different metadata.
Events from starknet will have different metadata than events from the API or from another chain.
But we cannot know which type of metadata it will be just based on the aggregate type itself. So this extra field will allow us to programmatically discriminate between this diversity of metadata schema.

It can be included in the json, but I think it's better to have it outside in order to know the target before deserialization.
It's also more high-level/domain-related than `block_number` or `transaction_hash`. Like `timestamp` it is common to all events, regardless of their nature/origin. For those reasons I think it's better to have it flattened.

`I think the concept of event metadata should only exist at the infrastructure layer.`

`StorableEvent` is defined in the domain. It contains everything to be stored with an event.
If we want to take `metadata` out of `StorableEvent` we change the `Store` trait
`	fn append(&self, aggregate_id: &A::Id, events: Vec<StorableEvent<A>>) -> Result<(), Error>;`
can become
`	fn append(&self, aggregate_id: &A::Id, events: Vec<(StorableEvent<A>, json::Value)>) -> Result<(), Error>;`

But this trait is still defined in the domain, so it changes nothing regarding your concern.

AnthonyBuisset (approved):

PR #264, opened 2022-10-04
🐛 remove contributor_id on unassign

ofux:


AnthonyBuisset (approved):

ofux:
> A e2e test would be great

yes, but not really doable until we can register users in e2e test ;)

AnthonyBuisset (approved):

PR #265, opened 2022-10-04
refactoring of the contributor aggregate module

AnthonyBuisset:
Standard proposal for domain code architecture, applied to the contributor aggregate


marketplace-core/src/application/contributor/refresh/test.rs
```diff
@@ -124,7 +124,7 @@ async fn refresh_contributors_from_events(
 	assert!(result.is_ok(), "{}", result.err().unwrap());
 
 	assert_eq!(
-		ContributorProjection {
+		ContributorDetails {
```
Bernardstanislas, line 127: I think `details` is a bit too generic, maybe `ContributorProfile`?

Bernardstanislas (approved):

PR #268, opened 2022-10-05
feat: add get_contributions route

tdelabro:


marketplace-core/src/dto/contributions.rs
```diff
@@ -29,3 +33,63 @@ impl ContributionCreation {
...
+				r#type: contribution.metadata.r#type,
+			},
+		}
+	}
+}
+
+pub fn build_contribution_dto(
+	contribution: ContributionProjection,
+	contributor_projection_repository: &Arc<dyn ContributorProjectionRepository>,
+) -> Option<Contribution> {
```
AnthonyBuisset, line 86: when can we have a`None` value ? 
If never I suggest to remove the `Option`, otherwise could be a good idea to log a warning as we might have corrupted data.

AnthonyBuisset (approved):
Very elegant solution for dynamic filtering !

marketplace-core/src/dto/contributions.rs
tdelabro: we can:
```rust
let contributor = contribution
		.contributor_id
		.clone()
		.and_then(|id| contributor_projection_repository.find_by_id(&id).ok());
```
Yes for the warning.

PR #275, opened 2022-10-07
🐛 fetch contributor from account instead of id

AnthonyBuisset:
As now the `contributor_account_address` column in `contributions` projection is used to references the contributor assigned, 
we need to fetch the contributor details from `contributors` projection `account` column instead of `id`.

e2e tests will be added in PR #268
https://github.com/onlydustxyz/marketplace-backend/pull/268/files#diff-2844ce2792781d70948ec9885d84e59f6e20fa1435ba0bf380d7969900423e91

### DIFF
--- a/marketplace-core/src/application/contributor/refresh/test.rs
+++ b/marketplace-core/src/application/contributor/refresh/test.rs
@@ -101,7 +101,6 @@ async fn refresh_contributors_from_events(
 	filled_database: Arc<DatabaseClient>,
 	mut github_client: MockGithubClient,
 	contributor_account_address: ContributorAccountAddress,
-	contributor_id: ContributorAccountAddress,
 ) {
 	let refresh_contributors_usecase: RefreshContributors = {
 		github_client.expect_find_user_by_id().returning(|_| Ok(Default::default()));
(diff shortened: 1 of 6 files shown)


marketplace-core/src/application/contributor/refresh/test.rs
```diff
@@ -119,13 +118,15 @@ async fn refresh_contributors_from_events(
...
+	let result = ContributorProjectionRepository::find_by_account_address(
+		&*filled_database.clone(),
+		&contributor_account_address,
+	);
 	assert!(result.is_ok(), "{}", result.err().unwrap());
 
 	assert_eq!(
 		ContributorProfile {
-			id: contributor_id,
+			id: contributor_account_address.clone(),
```
Bernardstanislas, line 129: Can we change the `ContributionProfile` to swap `id` with `account_address`?

marketplace-core/src/routes/contributors/list.rs
```diff
@@ -19,7 +19,7 @@ pub async fn get_contributor(
 	let contributor_id = contributor_id.into();
```
Bernardstanislas, line 19: `contributor_account_address`?

marketplace-core/src/application/contributor/refresh/test.rs
AnthonyBuisset: it will be done with [the clean-up task](https://www.notion.so/onlydust/Clean-up-contributor_id-from-codebase-d45efad1fede4b6f8de4751aadf233b1) when we remove the contributor_id from the `GithubAccountAssociated` event as well.

marketplace-core/src/routes/contributors/list.rs
AnthonyBuisset: this one has a FE impact, it will be removed when front-end uses the new `GET /contributors/contributor_account` route, again in [the cleanup task](https://www.notion.so/onlydust/Clean-up-contributor_id-from-codebase-d45efad1fede4b6f8de4751aadf233b1)

Bernardstanislas (approved):

PR #278, opened 2022-10-11
Add project id in contribution endpoint

lechinoix:


marketplace-core/src/routes/contributions/get.rs
```diff
@@ -4,36 +4,17 @@ use http_api_problem::HttpApiProblem;
...
-			ContributorAccountAddress::from(HexPrefixedString::from(address));
-		filters.push(contributor_account_address.into())
-	}
-	if let Some(project_id) = project_id {
-		filters.push(project_id.into())
-	}
-
-	let contributions =
-		contribution_repository.find(&filters).map_err(|e| e.to_http_api_problem())?;
+	let contributions = contribution_repository.find(&[]).map_err(|e| e.to_http_api_problem())?;
```
Bernardstanislas, line 17: Just checked and this is the only use of this method, so we might remove the filters from the signature and the codebase

marketplace-tests/Cargo.toml
```diff
@@ -7,7 +7,7 @@ version = "0.1.0"
 [lib]
 
 [features]
-default = []
+default = ["with_e2e_tests"]
```
AnthonyBuisset, line 10: Not sure it is a good idea as e2e tests will be triggered when running `cargo test` (instead of unit tests only)

tdelabro: yep I added it for the debuging and it stayed. It's probably the thing making the tests fail. 
Better remove it

lechinoix: On it thx !

marketplace-infrastructure/src/database/repositories/contribution.rs
```diff
@@ -6,27 +6,15 @@ use marketplace_domain::*;
...
-		for filter in filters.iter() {
-			match filter {
-				ContributionProjectionFilter::Project(id) =>
-					query = query.filter(dsl::project_id.eq(id.to_string())),
-				ContributionProjectionFilter::Contributor(id) =>
-					query = query.filter(dsl::contributor_id.eq(id.to_string())),
-			}
-		}
-
+	fn find(&self) -> Result<Vec<ContributionProjection>, ContributionProjectionRepositoryError> {
```
tdelabro, line 9: Du coup c'est plus vraiment un `find`.
C'est un `list_all`

marketplace-infrastructure/src/database/repositories/contribution.rs
```diff
@@ -6,27 +6,15 @@ use marketplace_domain::*;
...
 		let connection = self
 			.connection()
 			.map_err(|e| ContributionProjectionRepositoryError::Infrastructure(e.into()))?;
 
-		let contributions =
-			query.load::<models::Contribution>(&*connection).map_err(DatabaseError::from)?;
+		let contributions = dsl::contributions
+			.into_boxed()
```
tdelabro, line 15: T'as pas besoin du `into_boxed`.
C'est utilse que quand tu veux modifier ta query au runtime: https://docs.diesel.rs/diesel/query_dsl/trait.QueryDsl.html#method.into_boxed

tdelabro (dismissed):

PR #279, opened 2022-10-12
feat(indexer): add/fix event ContributionClosed and ContributionOpened

tdelabro:


marketplace-indexer/src/infrastructure/apibara/starknet/events/contribution/opened.rs
```diff
@@ -0,0 +1,63 @@
...
+mod test {
+	use std::str::FromStr;
+
+	use super::*;
+	use marketplace_domain::HexPrefixedString;
+	use rstest::*;
+
+	#[fixture]
+	fn apibara_event_data() -> Topics {
+		vec![vec![
```
AnthonyBuisset, line 32: should be an empty vec right ?

tdelabro: yup. you are right

PR #280, opened 2022-10-12
🚌 Event bus kick-off 🎉

AnthonyBuisset:
Add concept of event bus in application, based on RabbitMQ.
In this PR I am adding a new crate `marketplace-event-store`:
 - because the event store is unique, central and critical 
 - it might rely on its own infrastructure (database instance) in a near future 
 - it consumes events published from our aggregates and smart contracts
 - once persisted, it publishes the domain event to update our projections

I am adding as well some traits in `marketplace-domain/messaging` module:
 - The `Message` trait is what is being shared within a `Bus`
 - The `Publisher` trait to publish `Messages`
 - The `Subscriber` trait to subscribe to `Messages` and execute a user-defined callback
 - The `Bus` trait is both a `Publisher` and a `Subscriber` of `Messages`

There is also an implementation of the `Bus` in the `marketplace-infrastructure` crate that exchange domain events.
This implementation is meant to be used to update our projections.

More info [here](https://www.notion.so/onlydust/Event-bus-using-RabbitMQ-c4d36a9331d24cb28424958cd83bb00e)


marketplace-domain/src/messaging/bus.rs
```diff
@@ -0,0 +1,3 @@
+use super::{Message, Publisher, Subscriber};
+
+pub trait Bus<M: Message>: Publisher<M> + Subscriber<M> {}
```
Bernardstanislas, line 3: What disturbs me a bit is that in DDD in general as well as when we talked about it recently, we decided to have bounded contexts as subfolders of the `domain` crate. And here we are adding the `messaging` folder, which is not a bounded context. 
The way I used to deal with this was to put directly these cross-contexts objects at the root of the domain, here directly in `src`.

marketplace-infrastructure/src/event_bus/event_bus.rs
```diff
@@ -0,0 +1,180 @@
...
+	where
+		C: Fn(Event) -> F + Send + Sync,
+		F: Future<Output = Result<(), anyhow::Error>> + Send,
+	{
+		let queue = self
+			.channel
+			.queue_declare(
+				"",
+				QueueDeclareOptions {
+					exclusive: true, // only one consumer on this queue
```
Bernardstanislas: Amazing! What happens if the bin is scaled up by Heroku? (wont's happen in our current config)

marketplace-infrastructure/src/event_bus/mod.rs
```diff
@@ -1,2 +1,174 @@
-mod event_bus;
-pub use event_bus::EventBus;
+use anyhow::anyhow;
```
Bernardstanislas, line 1: It's weird Git does not show a file move

Procfile
```diff
@@ -1,3 +1,4 @@
 web: ROCKET_PORT=$PORT ./target/release/marketplace-core
 indexer: ./target/release/marketplace-indexer
```
Bernardstanislas, line 2: If I understand correctly, we will eventually have only one bin which will be a merge of `indexer` and `event-store` when we are confident enough to only use the event bus as the events transport?

marketplace-event-store/src/main.rs
```diff
@@ -0,0 +1,25 @@
...
+	dotenv().ok();
+	logger::set_default_global_logger().cancel_reset();
+
+	let event_store_bus = Bus::default().await?;
+
+	event_store_bus
+		.consume(|data| async move {
+			let event: Event = serde_json::from_slice(&data)?;
+			println!(
+				"[event-store] Received message: {}",
```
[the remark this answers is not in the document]
AnthonyBuisset: sure !

marketplace-infrastructure/src/event_bus/event_bus.rs
AnthonyBuisset: it returns an error and prevent the back-end to start

marketplace-infrastructure/src/event_bus/mod.rs
AnthonyBuisset: because I changed it a bit as well for a clippy warning

Procfile
AnthonyBuisset: We will have:
 - 1 binary for our API 
 - 1 binary for the event-listeners (each one with a dedicated queue)
 - 1 binary for the event store
 - 1 binary for the indexer (1 per ecosystem I guess, but we are not there yet)

marketplace-domain/src/messaging/bus.rs
AnthonyBuisset: As discussed I will split the Publisher and Subscriber implementation to be closer to where they are needed, as described here: https://www.notion.so/onlydust/Event-bus-using-RabbitMQ-c4d36a9331d24cb28424958cd83bb00e#6e5621603b55428fa22956b08da02416

marketplace-domain/src/messaging/subscriber.rs
```diff
@@ -0,0 +1,23 @@
...
+use anyhow::Result as AnyResult;
+use async_trait::async_trait;
+use std::future::Future;
+use thiserror::Error;
+
+#[derive(Debug, Error)]
+pub enum Error {
+	#[error("Something happened at infrastructure level")]
+	Infrastructure(#[source] anyhow::Error),
+	#[error(transparent)]
```
lechinoix, line 11: It is really hard for me to understand when we use transparent and when we use a hardcoded error message. How are those errors logged by the event bus ? Do we have the stacktrace from the original error ?

marketplace-event-store/src/event_bus.rs
```diff
@@ -0,0 +1,146 @@
...
+					durable: true,      // persist messages during restart
+					exclusive: true,    // only one consumer on this queue
+					auto_delete: false, // do not delete the queue at shutdown
+					..Default::default()
+				},
+				Default::default(),
+			)
+			.await
+			.map_err(|e| {
+				error!("Failed while declaring the queue: {e}");
```
lechinoix, line 64: I have the feeling that logging error and bubbling it as well could lead to duplicated error messages in the logs, but better too much logs than not enough

marketplace-infrastructure/src/event_bus/event_bus.rs
```diff
@@ -0,0 +1,180 @@
...
+	Serialize(#[from] serde_json::Error),
+}
+
+pub struct EventBus {
+	_connection: Connection,
+	channel: Channel,
+	exchange_name: &'static str,
+}
+
+impl EventBus {
```
lechinoix: This code also exists in the domain package no ? Is it tech debt or are we fine with this ?

marketplace-infrastructure/src/event_bus/mod.rs
```diff
@@ -28,7 +28,7 @@ impl EventBus {
 		channel
 			.exchange_declare(
 				exchange_name,
-				lapin::ExchangeKind::Topic,
+				lapin::ExchangeKind::Fanout,
```
lechinoix, line 31: Why this change ? Wasn't it what you draw on your incredible scheme ?

marketplace-infrastructure/src/event_bus/mod.rs
```diff
@@ -0,0 +1,174 @@
...
+	Amqp(#[from] lapin::Error),
+}
+
+pub struct EventBus {
+	_connection: Connection,
+	channel: Channel,
+	exchange_name: &'static str,
+}
+
+impl EventBus {
```
lechinoix, line 24: Isn't this file very similar to marketplace-event-store/src/event_bus.rs ?

marketplace-event-store/src/bus.rs
```diff
@@ -0,0 +1,116 @@
...
+pub enum Error {
+	#[error(transparent)]
+	Environment(#[from] VarError),
+	#[error(transparent)]
+	AMQP(#[from] lapin::Error),
+	#[error(transparent)]
+	Callback(#[from] anyhow::Error),
+	#[error("Could not deliver message to event store")]
+	Delivery,
+	#[error("Failed while serializing event")]
```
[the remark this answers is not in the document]
AnthonyBuisset: because there is no underlying error (no source) so nothing to be transparent
it is an error raised when the consumer returns `NACK` to the publisher

scripts/docker/dev/docker-compose.yml
```diff
@@ -68,6 +68,13 @@ services:
     environment:
       HTTPMOCK_MOCK_FILES_DIR: /mocks
 
+  rabbitmq-management:
```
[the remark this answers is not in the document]
AnthonyBuisset: indeed

marketplace-event-store/src/event_bus.rs
AnthonyBuisset: yes you are right, this is done in case the `?` on the line below (when returning the `NACK`) fails, not to lose the original error. But in most cases, there will be double logging indeed.

marketplace-domain/src/messaging/subscriber.rs
tdelabro: We respect the http api problem norm, through this crate https://docs.rs/http-api-problem/latest/http_api_problem/, to make front aware of errors.

In a nutshell, we can carry tree pieces of information:
- the status code 
- the error message
- the error details

the status code is set in the to_http_api_problem trait/file

the error message comes from something like this `	#[error("Something happened at infrastructure level")]`
It is a description of what happened from the service/repository point of view

The details are the errors our infra gave us, stringified via anyhow (GithubError or SqlError or StarknetError).

Everything else is transparent.

tdelabro: So yes, there is something odd there

marketplace-infrastructure/src/event_bus/mod.rs
AnthonyBuisset: yes very, but I did not find a good way to refactor it as some implementation details differ (one uses an exchange to allow multiple consumers, the other has an exclusive named queue)
It is a will of rabbitmq to write expressive code like this, to be integrated as a mdw

marketplace-domain/src/messaging/subscriber.rs
AnthonyBuisset: I do not think we have the whole stacktrace, those errors should be logged when raised, then returned up to the caller.
Here I mask infrastructure errors not to leak implementation details. The transparent for `Processing` type is because this specific error is raised by the callback of the message processing, which is our own code (see `subscribe` method implementation)

marketplace-infrastructure/src/event_bus/mod.rs
AnthonyBuisset: indeed, but it is not used yet. It would require to complexify a bit the `subscribe` function to pass a `topic`. So I decided to fallback to a `FanOut` type that is simpler.
It will be possible to change it back if needed in the future

lechinoix: Okay, it was to be sure I got it thanks !

marketplace-event-store/src/event_bus.rs
lechinoix: Okay, indeed it is not very relevant here. It happened to my project past year, we were logging errors at application / domain and infrastructure level and then it was a nightmare for monitoring, but it is a broader problem.

Bernardstanislas (approved):

PR #282, opened 2022-10-13
fix code cov on filters

lechinoix:
- Fix env.example to have e2e tests run locally
- Remove dead code
- Implement test for added debug log


marketplace-infrastructure/src/starknet_account_verifier/account_verifier.rs
```diff
@@ -69,3 +65,26 @@ impl OnChainAccountVerifier for StarkNetClient {
...
+
+	#[tokio::test]
+	async fn check_signature_error() {
+		let client = StarkNetClient::default();
+		let contributor_account_address = ContributorAccountAddress::from_str("0x1234").unwrap();
+		let signed_data = &SignedData {
+			hash: FieldElement::from_str("0x112233").unwrap(),
+			signature: Signature {
+				r: FieldElement::from_str("0x112233").unwrap(),
+				s: FieldElement::from_str("0x112233").unwrap(),
```
lechinoix, line 82: I do not know what are those value, could be worth to put production like data

AnthonyBuisset: https://en.wikipedia.org/wiki/Elliptic_Curve_Digital_Signature_Algorithm
Good luck with that 😁

AnthonyBuisset (approved):

marketplace-infrastructure/src/starknet_account_verifier/account_verifier.rs
lechinoix: Hahaha I believe @Bernardstanislas is somehow a pro about those

PR #283, opened 2022-10-15
refactor event bus implementation

AnthonyBuisset:
Refactoring of my previous PR#280
Following Stan's comment https://github.com/onlydustxyz/marketplace-backend/pull/280#discussion_r994345151

Now we have:
   - messaging module in domain with common behaviours (`Message`, `Destination`, `Publisher`, `Subscriber`)
   - amqp module in infrastructure with `lapin` logic and utils functions to declare queues and exchanges, publish and consume raw data
   - amqp module contains also a default implementation of the `Publisher` and `Subscriber` for a `Bus`
   - `event_bus` module in infrastructure to expose a function to easily create a consumer (we can have plenty of this one, all connected to the same echange, each with a dedicated queue)
   - `marketplace-event-store` now just uses infrastructure crate to create its own consumer (with a single queue, exclusive) and exposes its queue name for others to publish events to

Hope you like the new design !
Comments welcome, as always ;-)


marketplace-event-store/src/event_bus.rs
```diff
@@ -1,144 +1,29 @@
...
-
-		Ok(())
-	}
+use lapin::{options::QueueDeclareOptions, Connection};
+use log::info;
+use marketplace_infrastructure::amqp::{Error as EventBusError, EventBus};
+
+pub async fn consumer() -> Result<EventBus, EventBusError> {
+	let connection = Connection::connect(&amqp_address()?, Default::default()).await?;
+	info!("🔗 Event store connected");
```
Bernardstanislas: I find this message a bit confusing, it can mean that the app is connected to the event store, whereas you want to say that the events are getting listened to and stored

marketplace-infrastructure/src/amqp/mod.rs
```diff
@@ -91,12 +91,12 @@ impl EventBus {
 		}
 	}
 
-	async fn publish(&self, data: &[u8]) -> Result<Confirmation, Error> {
+	async fn publish(&self, routing_key: &str, data: &[u8]) -> Result<Confirmation, Error> {
```
Bernardstanislas: Why do you sometimes call this the `routing_key`, and sometimes the `destination`?

marketplace-indexer/src/lib.rs
```diff
@@ -16,9 +15,9 @@ pub async fn main() {
 	github::Client::initialize();
 
 	let database = Arc::new(database::Client::new(database::init_pool()));
-	let event_store_bus = event_store_bus::publisher()
-		.await
-		.expect("Unable to connect to the event store bus");
+	let event_store_bus =
```
Bernardstanislas: I would name it `event_bus`, I think the indexer shouldn't know the specifics of how events are used after it sends them in the bus

marketplace-infrastructure/src/amqp/mod.rs
```diff
@@ -19,14 +19,14 @@ pub enum Error {
 	Amqp(#[from] lapin::Error),
 }
 
-pub struct EventBus {
+pub struct Bus {
```
Bernardstanislas: I find it a bit too generic, we are talking here about a `MessageBus`

marketplace-indexer/src/lib.rs
```diff
@@ -7,16 +7,15 @@ use crate::{domain::*, infrastructure::apibara};
...
 use std::sync::Arc;
 
 pub async fn main() {
 	dotenv().ok();
 	github::Client::initialize();
 
 	let database = Arc::new(database::Client::new(database::init_pool()));
-	let event_store_bus =
-		EventBus::default().await.expect("Unable to connect to the event store bus");
+	let event_store_bus = Bus::default().await.expect("Unable to connect to the event store bus");
```
Bernardstanislas, line 18: As I said in another comment, I think we should avoid talking about the event store in the indexer, it should just know it sends events in a message bus

marketplace-infrastructure/src/amqp/event_bus.rs
```diff
@@ -1,13 +1,11 @@
-use super::{Bus, Error};
+use super::{Bus, ConsumableBus, Error};
 use lapin::options::QueueDeclareOptions;
 use log::info;
 
 pub const EXCHANGE_NAME: &str = "events";
 
-pub async fn consumer() -> Result<Bus, Error> {
+pub async fn consumer() -> Result<ConsumableBus, Error> {
 	let event_bus = Bus::default()
```
Bernardstanislas: Variable name should not hold the notion of event anymore

marketplace-infrastructure/src/amqp/event_bus.rs
```diff
@@ -17,8 +15,9 @@ pub async fn consumer() -> Result<Bus, Error> {
 			},
 		)
 		.await?
-		.binded()
+		.with_exchange(EXCHANGE_NAME)
 		.await?;
+
 	info!("🔗 Event bus connected");
```
Bernardstanislas: Same here, there is no trace of messages being events anymore

marketplace-infrastructure/src/amqp/mod.rs
```diff
@@ -43,37 +39,63 @@ impl Bus {
...
 	}
+}
 
-	pub async fn with_queue(
-		self,
-		queue_name: &'static str,
-		options: QueueDeclareOptions,
-	) -> Result<Self, Error> {
-		self.channel.queue_declare(queue_name, options, Default::default()).await?;
+pub struct ConsumableBus {
```
Bernardstanislas: What is the goal of distinguishing the `Bus` from the `ConsumableBus`? I think we can expect a `MessageBus` to be both publishable et consumable, these two features being the minimal coherent set of features of a bus?
Splitting them in two different concepts makes these concepts harder to understand IMHO

marketplace-event-store/src/lib.rs
```diff
@@ -10,11 +10,11 @@ use marketplace_infrastructure::amqp::{event_bus::EXCHANGE_NAME, Bus};
 use std::sync::Arc;
 
 pub async fn main() -> Result<()> {
-	let event_store_bus = event_bus::consumer().await?;
-	let event_bus = Arc::new(Bus::default().await?);
+	let input_bus = bus::consumer().await?;
```
Bernardstanislas, line 13: I like the idea, I would even go further because the event store does know what to expect from these two message buses.
I would go for `inbound_event_bus` and `outbound_event_bus`, WDYT?

marketplace-event-store/src/event_bus.rs
AnthonyBuisset: yes you're right, I will review the logging in general and improve it

marketplace-infrastructure/src/amqp/mod.rs
AnthonyBuisset: routing key and exchange name are amqp information
destination is a notion of the domain (destination being in that case a queue or an exchange)

marketplace-infrastructure/src/amqp/mod.rs
AnthonyBuisset: Actually, the bus is transmitting raw data (`&[u8]`) that is why I did not name it `MessageBus`. Also the `Message` is in the domain, but not used at all here in this infrastructure file.

marketplace-infrastructure/src/amqp/event_bus.rs
AnthonyBuisset: agreed

marketplace-infrastructure/src/amqp/event_bus.rs
AnthonyBuisset: good point

marketplace-infrastructure/src/amqp/mod.rs
AnthonyBuisset: The Bus can publish to a destination
To be able to consume messages from it, we need at least to declare a queue (and optionally an exchange).
And to declare an exchange, we need to declare a queue first.
The original design had `queue_name` and `exchange_name` as attributes that were empty in case of a simple publisher, which is 💩 

The API is :
```
Bus::default() -> Bus (can publish)
Bus::default().with_queue() -> ConsumableBus (can consume)
Bus::default().with_queue().with_exchange() -> ConsumableBus (can consume)
```

marketplace-event-store/src/lib.rs
AnthonyBuisset: I like it

Bernardstanislas (approved):

PR #285, opened 2022-10-18
Use caller address in event decoding

tdelabro:


marketplace-indexer/src/infrastructure/apibara/starknet/indexing/block.rs
```diff
@@ -68,14 +73,37 @@ impl AsEvents for Block {
 	}
 }
 
+fn find_event_caller(transactions: &[Transaction], transaction_hash: &Vec<u8>) -> Option<Vec<u8>> {
+	transactions.iter().find_map(|t| {
+		if let transaction::Transaction::Invoke(invoke_transaction) =
+			t.transaction.as_ref().expect("Should be present exist")
```
AnthonyBuisset, line 79: Error message to be rephrased (or changed to unwrap)

marketplace-indexer/src/infrastructure/apibara/starknet/events/mod.rs
```diff
@@ -37,6 +37,7 @@ pub struct Event {
 	pub transaction_hash: HexPrefixedString,
 	pub index: usize,
 	pub from_address: ContractAddress,
+	pub caller: Option<ContractAddress>,
```
AnthonyBuisset, line 41: could you please add it in the metadata as well ?

AnthonyBuisset (approved):

marketplace-indexer/src/infrastructure/apibara/starknet/indexing/block.rs
tdelabro: done

marketplace-indexer/src/infrastructure/apibara/starknet/events/mod.rs
tdelabro: done

marketplace-indexer/src/infrastructure/apibara/starknet/indexing/block.rs
```diff
@@ -68,14 +73,37 @@ impl AsEvents for Block {
...
+		{
+			let hash = invoke_transaction.common.as_ref().unwrap().hash.clone();
```
Bernardstanislas, line 81: is it ok to unwrap here?

Bernardstanislas: Also can you pattern match? It will be more elegant than this `if else` sequence

PR #286, opened 2022-10-18
move discord handle from contact information to contributor profile

AnthonyBuisset:
- 🗃️ migrate discord_handle to contributors table
- ⚰️ remove contact information code
- ♻️ change contributor insert to upsert when associating github user
- ✨ add discord handle in contributor aggregate and event
- ✨ upsert discord information upon event
- ♻️ make github id and user optional in contributor profile
- ✨ add register discord handle usecase
- ✨ return discord_handle in GET /contributor


marketplace-core/src/application/contributor/register_discord_handle.rs
```diff
@@ -0,0 +1,85 @@
...
+				if let Event::Contributor(contribution_event) = event {
+					StorableEvent {
+						deduplication_id: self.uuid_generator.new_uuid().to_string(),
+						event: contribution_event.clone(),
+						timestamp: Utc::now().naive_utc(),
+						origin: EventOrigin::BACKEND,
+						metadata: Default::default(),
+					}
+				} else {
+					panic!("Contribution event expected");
```
Bernardstanislas, line 72: Why so? And I think it's `Contributor` here

Bernardstanislas (approved):
I think this PR could have been split between removing the contact information table, and event-sourcing the discord registration.
We also need to coordinate with the frontend because this introduces some breaking changes in the API

AnthonyBuisset:
Yes you're right. I mixed removing the table and removing the route with the same name. But it could have totally been split.

PR #296, opened 2022-10-25
Stop indexer in case of reorg

lechinoix:
Synthetic test created : https://app.datadoghq.eu/synthetics/details/ar8-vsz-9zj

### DIFF
--- a/marketplace-indexer/src/domain/observers/indexer.rs
+++ b/marketplace-indexer/src/domain/observers/indexer.rs
@@ -1,6 +1,6 @@
 use crate::domain::{IndexerService, Observer};
 use async_trait::async_trait;
-use log::{error, info};
+use log::{error, warn};
 use std::sync::Arc;
 
 pub struct IndexerObserver {
@@ -18,7 +18,7 @@ impl Observer for IndexerObserver {
 	async fn on_reorg(&self) {
 		let result = self.indexer_service.stop_indexer().await;
 		match result {
-			Ok(()) => info!("Indexer stopped"),
+			Ok(()) => warn!("[dd-id=indexer-stopped] Indexer stopped due to Reorg"),
 			Err(e) => error!("Failed to stop indexer: {}", e),
 		}
 	}
--- a/marketplace-indexer/src/domain/observers/indexer.rs
+++ b/marketplace-indexer/src/domain/observers/indexer.rs
@@ -18,7 +18,7 @@ impl Observer for IndexerObserver {
 	async fn on_reorg(&self) {
 		let result = self.indexer_service.stop_indexer().await;
 		match result {
-			Ok(()) => warn!("[dd-id=indexer-stopped] Indexer stopped due to Reorg"),
+			Ok(()) => warn!("Indexer stopped due to Reorg"),
 			Err(e) => error!("Failed to stop indexer: {}", e),
 		}
 	}
--- a/marketplace-indexer/src/domain/observers/indexer.rs
+++ b/marketplace-indexer/src/domain/observers/indexer.rs
@@ -1,0 +1,41 @@
+use crate::domain::{IndexerService, Observer};
+use async_trait::async_trait;
+use log::{error, info};
+use std::sync::Arc;
+
+pub struct IndexerObserver {
+	indexer_service: Arc<dyn IndexerService>,
+}
+
+impl IndexerObserver {
+	pub fn new(indexer_service: Arc<dyn IndexerService>) -> Self {
+		Self { indexer_service }
+	}
+}
+
+#[async_trait]
+impl Observer for IndexerObserver {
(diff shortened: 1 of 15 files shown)


lechinoix:
Todo :

- Rename ReorgObserver to IndexerProcessObserver
- Add env variable in Heroku
- Add an alert on indexer stopped

marketplace-indexer/src/domain/observers/indexer.rs
```diff
@@ -18,7 +18,7 @@ impl Observer for IndexerObserver {
 	async fn on_reorg(&self) {
 		let result = self.indexer_service.stop_indexer().await;
 		match result {
-			Ok(()) => info!("Indexer stopped"),
+			Ok(()) => warn!("[dd-id=indexer-stopped] Indexer stopped due to Reorg"),
```
Bernardstanislas, line 21: Is it a new convention? `[dd-id=indexer-stopped]`
Do use `tracing` instead, to use metadata instead of text logging

marketplace-indexer/src/infrastructure/heroku/indexer.rs
```diff
@@ -5,48 +5,55 @@ use serde_json::json;
...
+mod test {
+	use super::*;
+	use assert_matches::assert_matches;
+	use envtestkit::lock::lock_test;
+	use mockito::Matcher;
+
+	#[allow(clippy::await_holding_lock)]
+	#[tokio::test]
+	async fn http_call_fail() {
+		let _m = mockito::mock(
```
Bernardstanislas, line 39: ❤️

marketplace-indexer/src/infrastructure/heroku/indexer.rs
```diff
@@ -0,0 +1,52 @@
...
+impl IndexerService for HerokuClient {
+	async fn stop_indexer(&self) -> Result<(), anyhow::Error> {
+		self.scale_formation(INDEXER_FORMATION_TYPE, 0).await?;
+
+		Ok(())
+	}
+}
+
+fn heroku_token() -> String {
+	std::env::var("HEROKU_TOKEN").expect("HEROKU_TOKEN var must be set")
```
AnthonyBuisset: we should avoid expect here as we do not control when this function is called. Best would be to return a Result ans use `map_err` to add some context (like the env var name)

marketplace-indexer/src/infrastructure/heroku/mod.rs
```diff
@@ -0,0 +1,3 @@
+mod client;
+mod indexer;
+pub use indexer::*;
```
AnthonyBuisset, line 3: I do not think you need to export * here, only the `HerokuClient` should be enough

marketplace-indexer/src/infrastructure/heroku/client.rs
```diff
@@ -3,15 +3,31 @@ use reqwest::{self, header, Body};
...
+}
+
+pub struct HerokuClient {}
+
+impl HerokuClient {
+	pub fn new() -> Self {
+		Self {}
+	}
+
+	pub fn create_client(&self) -> Result<reqwest::Client, anyhow::Error> {
```
AnthonyBuisset, line 30: why not creating the client in the `new` above and keeping it as an attribute ? do we need to create a client at runtime ?

lechinoix: Hesitated a bit on this one, I was wondering if having a constructor that can possibly fail was a good practice or not

lechinoix: But yes actually now that I try again, it makes me manage error in the main process instead of the IndexerServiceObserver

marketplace-indexer/src/infrastructure/heroku/indexer.rs
lechinoix: Done !

marketplace-indexer/src/infrastructure/heroku/client.rs
```diff
@@ -20,14 +20,12 @@ impl HerokuEndpoint {
...
+	http_client: reqwest::Client,
+}
 
 impl HerokuClient {
-	pub fn new() -> Self {
-		Self {}
-	}
-
-	pub fn create_client(&self) -> Result<reqwest::Client, anyhow::Error> {
+	pub fn new() -> Result<Self, anyhow::Error> {
```
AnthonyBuisset, line 28: you can also use anyhow::Result

AnthonyBuisset (approved):

marketplace-indexer/src/infrastructure/heroku/client.rs
lechinoix: Ooooooo nice !!

PR #301, opened 2022-11-03
use typed aggregate event in aggregate modules

AnthonyBuisset:
- Using Aggregate::Event in domain aggregates for event sourcing
- Serializing Aggregate::Event in event-store
- Keeping DomainEvent for now for event bus and projectors

### DIFF
--- a/marketplace-core/src/lib.rs
+++ b/marketplace-core/src/lib.rs
@@ -188,10 +188,10 @@ pub async fn event_listeners_main() -> Result<()> {
 
 	event_consumer
 		.subscribe(|event: Event| async move {
-			debug!(
-				"[events] 📨 Received event: {}",
-				serde_json::to_string_pretty(&event)?
-			);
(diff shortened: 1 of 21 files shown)


marketplace-core/src/lib.rs
```diff
@@ -188,10 +188,10 @@ pub async fn event_listeners_main() -> Result<()> {
 
 	event_consumer
 		.subscribe(|event: Event| async move {
-			debug!(
-				"[events] 📨 Received event: {}",
-				serde_json::to_string_pretty(&event)?
-			);
+			if let Ok(event) = serde_json::to_string_pretty(&event) {
```
lechinoix, line 191: What is the case when it can fail ? Shouldn't we at least log the fail ?

lechinoix (approved):
Nice one !

marketplace-core/src/lib.rs
AnthonyBuisset: This is fixed in my next PR :-)

PR #302, opened 2022-11-03
project events from bus

AnthonyBuisset:


marketplace-core/src/event_listeners/mod.rs
```diff
@@ -1,20 +1,10 @@
...
-			}
+mod logger;
 
-			Ok(())
-		})
-		.await?;
+pub async fn main() -> Result<()> {
+	logger::spawn(event_bus::consumer().await?).await?;
 
 	Ok(())
```
Bernardstanislas, line 27: Hum what happens if the logger exits gracefully? I'm not sure it will be restarted by Heroku

marketplace-core/src/application/contribution/apply.rs
```diff
@@ -80,18 +70,13 @@ impl Usecase for ApplyToContribution {
 				metadata: Default::default(),
 			})
 			.collect();
+
 		self.event_publisher
 			.publish_many(
```
Bernardstanislas, line 75: I am not sure about where it should live but it would be great not to have to convert events into storable_events in the future.
Maybe the publisher is the good place to do it?

Bernardstanislas (approved):

marketplace-core/src/application/contribution/apply.rs
AnthonyBuisset: Indeed, we were discussing it with Olivier, several ways to improve this serialization/deserialization process

marketplace-core/src/event_listeners/mod.rs
AnthonyBuisset: I think this should never happen, but I will perform some tests

PR #304, opened 2022-11-10
improve event bus implementation

AnthonyBuisset:
- ⚰️ remove unused routing key from destination
- 🔒️ make process panic when at least one listener crashed
- 🐛 make queues and exchanges durable
- ♻️ better naming for consumer queues


ofux (approved):
LGTM.
We'll have to improve what happens in case of error (eg. when an event fails to be processed) in another POUR.

PR #309, opened 2022-11-10
🗃️ create payments projection table

AnthonyBuisset:


marketplace-infrastructure/src/database/schema.rs
```diff
@@ -70,6 +70,17 @@ diesel::table! {
     }
 }
 
+diesel::table! {
+    payments (id) {
+        id -> Uuid,
+        amount -> Nullable<Numeric>,
+        currency_code -> Nullable<Text>,
+        recipient_id -> Nullable<Uuid>,
+        reason -> Nullable<Jsonb>,
```
ofux, line 79: Why those columns are Nullable?

ofux (changes requested):
I think some columns should better be non-nullable. WDYT?

marketplace-infrastructure/src/database/schema.rs
AnthonyBuisset: because I forgot 🙈

ofux (approved):

PR #312, opened 2022-11-10
create mutation for emitting an event upon payment processed

AnthonyBuisset:
- ♻️ move graphql definitions in dedicated module
- ✨ create graphql mutation root object
- ✨ add Payment::mark_as_processed command
- ✨ introduce graphql context to inject dependencies
- ✨ create the eth_payment_processed mutation
- ♻️ make graphql endpoint async

### DIFF
--- a/marketplace-core/src/application/mod.rs
+++ b/marketplace-core/src/application/mod.rs
@@ -9,3 +9,5 @@ pub use project::*;
 
 mod contributor;
 pub use contributor::*;
+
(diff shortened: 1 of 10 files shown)


Bernardstanislas (commented):
Ok but let's not forget DDD :)
GraphQL is only our presentation layer, create a use case for what you're doing here

Bernardstanislas (approved):

PR #314, opened 2022-11-14
Deploy hasura

lechinoix:


hasura/migrations/Marketplace/1668425050572_init/up.sql
```diff
@@ -0,0 +1,144 @@
+SET check_function_bodies = false;
```
AnthonyBuisset, line 1: does this mean we are migrating now to hasura schemas ? how to transition from diesel ?

AnthonyBuisset (approved):

hasura/migrations/Marketplace/1668425050572_init/up.sql
lechinoix: Hey Antho ! Hasura generated this automatically, I'll remove it for now and we'll add it back later

PR #318, opened 2022-11-15
Install Hasura auth

lechinoix:
- Add remote schema permissions
- Update hasura mapping
- Add hasura auth in local


AnthonyBuisset (approved):

scripts/docker/dev/docker-compose.yml
```diff
@@ -45,36 +36,57 @@ services:
...
+      AUTH_SMTP_HOST: mailhog
+      AUTH_SMTP_PORT: 1025
+      HASURA_GRAPHQL_JWT_SECRET: '{"type":"HS256","key":"5152fa850c02dc222631cca898ed1485821a70912a6e3649c49076912daa3b62182ba013315915d64f40cddfbb8b58eb5bd11ba225336a6af45bbae07ca873f3","issuer":"hasura-auth"}'
+      HASURA_GRAPHQL_ADMIN_SECRET: myadminsecretkey
+    ports:
+      - "4000:4000"
+    volumes:
+      - ./hasura/auth/email-templates:/app/email-templates
+
+  mailhog:
```
Bernardstanislas, line 73: I'm not sure we'll use it

scripts/docker/dev/docker-compose.yml
```diff
@@ -45,36 +36,57 @@ services:
...
       HASURA_GRAPHQL_ADMIN_SECRET: myadminsecretkey
       HASURA_GRAPHQL_UNAUTHORIZED_ROLE: public
 
+  hasura-auth:
+    image: nhost/hasura-auth:latest
+    depends_on:
+      - postgres
+      - graphql-engine
+    env_file:
+      - ../../../hasura/auth/config.env
```
Bernardstanislas, line 60: can we use the repo's `.env` file instead? I think there is no conflict of variables and it would help having all the stack config in the same place

lechinoix: I had the feeling that it was more a configuration file that does not depend on the environment, and therefore I believe we should commit it to avoid having different configurations on different envs, what do you think ?

scripts/docker/dev/docker-compose.yml
lechinoix: Okay, I'll remove it then :)

PR #320, opened 2022-11-15
feat: add project_leads table, change types, rename

tdelabro:


marketplace-domain/src/project/id.rs
```diff
@@ -1,3 +1 @@
-use crate::GithubProjectId;
-
-pub type Id = GithubProjectId;
+pub type Id = uuid::Uuid;
```
AnthonyBuisset, line 1: we should use the [new type pattern](https://www.worthe-it.co.za/blog/2020-10-31-newtype-pattern-in-rust.html), using [derive_more](https://crates.io/crates/derive_more) crate it is very convenient.

marketplace-domain/src/project/mod.rs
```diff
@@ -1,5 +1,5 @@
 mod projections;
-pub use projections::LeadContributor as LeadContributorProjection;
+pub use projections::ProjectLead as LeadContributorProjection;
```
AnthonyBuisset, line 2: alias to be renamed as well

marketplace-infrastructure/src/database/schema.rs
```diff
@@ -26,21 +24,14 @@ diesel::table! {
...
 
-diesel::table! {
-    lead_contributors (project_id, contributor_id) {
-        project_id -> Text,
-        contributor_id -> Uuid,
-    }
-}
-
-diesel::table! {
+table! {
```
AnthonyBuisset, line 34: you should upgrade your diesel CLI to avoid changes like that

marketplace-infrastructure/src/database/models/lead_contributors.rs
```diff
@@ -4,7 +4,7 @@ use uuid::Uuid;
 
 #[derive(Queryable, Debug, Serialize, Deserialize, Insertable)]
 #[serde(crate = "rocket::serde")]
-pub struct LeadContributor {
-	pub project_id: String,
-	pub contributor_id: Uuid,
+pub struct ProjectLead {
+	pub project_id: Uuid,
```
lechinoix, line 8: Shouldn't you use the ProjectId type here ? Or maybe now it is pretty simple just delete the ProjectId type ?

tdelabro: Here we are defining an object that has to be insertable/queryable by diesel. This means each one of the types it contains should implement fromSqlType/toSqlType (probably not the actual names) traits.
Here I can indeed use `ProjectId` as it is a simple alias. But if I use the newtype pattern instead as recommended by Anthony, it won't work anymore.

The rule is more or less, when writing those db models object, use the native rust types as much as possible

marketplace-domain/src/project/id.rs
tdelabro: Super crate ! Je connaissais pas. Merci

marketplace-domain/src/project/mod.rs
tdelabro: done.
J'abandonne le postfix "Projection" par contre.
On avait dit qu'on pouvait se permettre se genre de simplifications

marketplace-infrastructure/src/database/schema.rs
tdelabro: done

lechinoix (approved):

migrations/2022-11-16-061721_add-payment-requests-table/up.sql
```diff
@@ -0,0 +1,9 @@
+CREATE TABLE payment_requests(
+    id UUID PRIMARY KEY,
+    project_id UUID NOT NULL REFERENCES projects(id),
+    requestor_id UUID NOT NULL REFERENCES users(id),
+    recipient_id UUID NOT NULL REFERENCES users(id),
+    amount_in_usd INT NOT NULL,
```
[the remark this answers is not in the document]
tdelabro: Dac!

marketplace-domain/src/payment_request/id.rs
```diff
@@ -0,0 +1,8 @@
+use derive_more::{AsRef, Display, From, Into};
+use serde::{Deserialize, Serialize};
+use uuid::Uuid;
+
+#[derive(
+	Debug, Clone, Copy, Default, Serialize, Deserialize, PartialEq, Eq, Display, From, Into, AsRef,
```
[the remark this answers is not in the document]
tdelabro: I think, From is wrapped => wrapper and Into is wrapper => wrapped. But I have to double-check

marketplace-domain/src/project/mod.rs
AnthonyBuisset: fine for me, especially for projections

PR #321, opened 2022-11-15
Improve AMQP and event store error handling

ofux:
- :recycle: Improve subscriber error handling
- :recycle: in case of duplicated events, the event store repository logs an error but returns Ok to simply ignore the event
- :white_check_mark: add some infrastructure tests for amqp

### DIFF
--- a/marketplace-domain/src/messaging/subscriber.rs
+++ b/marketplace-domain/src/messaging/subscriber.rs
@@ -19,14 +19,15 @@ pub enum Error {
 
 #[derive(Debug, Error)]
 pub enum CallbackError {
-	// Returning a BadMessage error will discard the current message. It won't be requeued.
-	#[error("Invalid message")]
-	BadMessage(#[source] anyhow::Error),
+	// Returning a Discard error will discard the current message and process the next one. It
+	// won't be requeued.
+	#[error("Ignoring message")]
+	Discard(#[source] anyhow::Error),
 
-	// Returning an InternalError error will stop the consuming of messages. The current message
-	// will be automatically requeued.
-	#[error("Internal error while processing the message")]
-	InternalError(#[from] anyhow::Error),
+	// Returning an Fatal error will stop the consuming of messages. The current message
+	// must be automatically requeued by the message broker.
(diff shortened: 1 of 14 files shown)


marketplace-domain/src/messaging/subscriber.rs
```diff
@@ -18,10 +17,22 @@ pub enum Error {
 	Deserialize(#[from] serde_json::Error),
 }
 
+#[derive(Debug, Error)]
+pub enum CallbackError {
+	// Returning a BadMessage error will discard the current message. It won't be requeued.
+	#[error("Invalid message")]
+	BadMessage(#[source] anyhow::Error),
```
AnthonyBuisset, line 24: shouldn't we have more expressive values ? like `Discard`, `Retry`, etc

marketplace-infrastructure/src/amqp/subscriber.rs
```diff
@@ -74,3 +74,200 @@ impl ConsumableBus {
...
+			.await
+			.unwrap()
+			.publish(&*String::new(), queue_name, "bad-message".as_bytes())
+			.await
+			.unwrap();
+
+		assert!(!confirmation.is_nack());
+	}
+
+	fn assert_error_message(error: SubscriberError, expected_error_message: &str) {
```
AnthonyBuisset, line 145: `assert_processing_error_message` ?

AnthonyBuisset (approved):

marketplace-domain/src/messaging/subscriber.rs
ofux: yes, good idea

ofux: WDYT of `Discard` and `Fatal`? Because the `Fatal` one actually terminates message consuming. Then, the message broker will requeue it so it will be retried, but that's not directly done by our code.

AnthonyBuisset: yes fine for me

AnthonyBuisset (approved):

PR #323, opened 2022-11-15
Add permissions to publicly read model

lechinoix:


hasura/metadata/databases/default/tables/public_users.yaml
```diff
@@ -16,3 +16,12 @@ object_relationships:
         remote_table:
           name: users
           schema: auth
+select_permissions:
+  - role: public
+    permission:
+      columns:
+        - discord_handle
```
AnthonyBuisset, line 23: are we sure we want to expose contact information publicly ?

AnthonyBuisset (approved):

PR #324, opened 2022-11-15
Plug backend to Hasura

lechinoix:


hasura/metadata/remote_schemas.yaml
```diff
@@ -1 +1,28 @@
-[]
+- name: Backend
+  definition:
+    url_from_env: BACKEND_GRAPHQL_URL
+    timeout_seconds: 60
+    customization: {}
+  permissions:
+    - role: project_lead
```
AnthonyBuisset, line 7: this one is to mark the payment as processed, so by the payer role and not the project lead

lechinoix: Ah alright well seen !

AnthonyBuisset (approved):

PR #332, opened 2022-11-17
Mutation create payment request

tdelabro:


marketplace-core/src/routes/payment_request/mod.rs
```diff
@@ -0,0 +1,66 @@
...
+		.into(),
+		deduplication_id: uuid_generator.new_uuid().to_string(),
+		timestamp: Utc::now().naive_utc(),
+		metadata: serde_json::Value::default(),
+		origin: EventOrigin::BACKEND,
+	};
+
+	event_publisher
+		.publish(
+			Destination::Queue(String::from(QUEUE_NAME)),
```
ofux, line 59: not directly related to this PR, but this constant should really be renamed to `EVENT_STORE_QUEUE` or equivalent

marketplace-domain/src/project/events.rs
```diff
@@ -1,8 +1,23 @@
...
+pub enum Event {
+	Created {
+		project_id: ProjectId,
+	},
+	PaymentRequested {
+		id: PaymentRequestId,
+		project_id: ProjectId,
+		requestor_id: UserId,
+		recipient_id: UserId,
+		amount_in_usd: u32,
```
ofux, line 17: I thought (but might be wrong) that we wanted to use a library like [rusty_money](https://docs.rs/rusty-money/latest/rusty_money/) to represent payment amounts? Was it discarded or postponed?

Of course, this doesn't necessarily apply to DTOs, but I guess it would make sense in the domain (but again, maybe we postponed it?).

ofux (approved):
Approved with 2 comments

marketplace-domain/src/project/events.rs
AnthonyBuisset: for payments yes, but here it the amount is just informational, so we decided to stay with u32

marketplace-core/src/routes/payment_request/mod.rs
AnthonyBuisset: it was designed to be used as 
```
use event_store::bus::QUEUE_NAME as EVENT_STORE_QUEUE;
```

tdelabro: Yep I saw this. It add this copy pasting everytime we use it.
I had the idea to rename it `EVENT_STORE_QUEUE` in another PR too
Coz if we have to rename it `EVENT_STORE_QUEUE` each time we import it, we should export it as `EVENT_STORE_QUEUE`

ofux:
This feature was finally reimplemented as a mutation: https://github.com/onlydustxyz/marketplace-backend/pull/350
Was re-done from scratch as a lot of refactoring occurred in the mean time.

PR #333, opened 2022-11-17
:card_file_box: Remove our users

Bernardstanislas:


lechinoix:
Might be cool to fix this in the same MR no ?

hasura/metadata/databases/default/tables/auth_users.yaml
```diff
@@ -91,15 +91,6 @@ object_relationships:
...
       foreign_key_constraint_on: default_role
-  - name: user_info
-    using:
-      manual_configuration:
-        column_mapping:
-          id: id
-        insertion_order: null
-        remote_table:
-          name: users
-          schema: public
```
lechinoix, line 102: I think you'll also have to remove relations based on the foreign_keys you removed, like in `public_project_leads`

lechinoix (approved):

PR #337, opened 2022-11-18
Add inconsistencies check in CI

lechinoix:
- Fix Makefile
- Add inconsistencies check in CI

### DIFF
--- a/.github/actions/env-setup/action.yml
+++ b/.github/actions/env-setup/action.yml
@@ -8,6 +8,8 @@ runs:
     - name: Run docker-compose
       shell: bash
       run: docker-compose up -d
+      env:
+        BACKEND_GRAPHQL_URL: https://staging.api.onlydust.xyz/graphql
 
     - name: Install diesel CLI
       shell: bash
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -99,6 +99,37 @@ jobs:
           token: ${{ secrets.GITHUB_TOKEN }}
           args: --bins --tests --examples -- -D warnings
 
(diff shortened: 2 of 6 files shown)


.github/actions/env-setup/action.yml
```diff
@@ -7,7 +7,9 @@ runs:
   steps:
     - name: Run docker-compose
       shell: bash
-      run: docker-compose -f ./scripts/docker/dev/docker-compose.yml up -d
+      run: docker-compose up -d
+      env:
+        BACKEND_GRAPHQL_URL: https://staging.api.onlydust.xyz/graphql
```
ofux, line 12: The CI targets staging? Isn't it a bit dangerous for our staging env?

ofux (commented):
LGTM, but I didn't approve as I think targeting the staging env from our CI might be a bit dangerous. WDYT?

.github/actions/env-setup/action.yml
AnthonyBuisset: It will be when adding tests. We need to target local for the CI but there are difficulties to be tackled before.

lechinoix: Looks fine for me at this time, it will only pull the graphQL schema. But yeah it is kind of a hack, and we will need to change it as soon as we have e2e tests setup

lechinoix: And it will slow down this check a lot to compile and run the server in the job, with no added value

ofux: Alright, sounds good for now :)

ofux (approved):

PR #344, opened 2022-11-21
Project lead mutations

Bernardstanislas:


api/src/graphql/context.rs
```diff
@@ -4,7 +4,8 @@ use event_store::Event;
 use std::sync::Arc;
 
 pub struct Context {
-	pub create_payment_usecase: application::payment::create::Usecase,
+	pub mark_payment_as_processed_usecase: application::payment::create::Usecase,
```
AnthonyBuisset: merge issue with the variable name I presume

AnthonyBuisset (approved):

api/src/graphql/context.rs
Bernardstanislas: yesss :/

Bernardstanislas: Fixed in a later commit though

api/src/graphql/mutation.rs
```diff
@@ -47,4 +47,23 @@ impl Mutation {
...
+		project_id: Uuid,
+		leader_id: Uuid,
+	) -> FieldResult<Uuid> {
+		context
+			.assign_project_lead_usecase
+			.assign_leader(project_id.into(), leader_id.into())
+			.await?;
+
+		Ok(project_id)
+	}
```
ofux, line 68: Looks like mutations are a pass-through to usecases. Would it be possible to avoid the need to write this code? Not high prio, but maybe another good candidate for a macro?

domain/src/project/aggregate.rs
```diff
@@ -19,3 +29,110 @@ impl From<ProjectEvent> for Event {
...
+	fn apply_event(mut self, event: &Self::Event) -> Self {
+		match event {
+			ProjectEvent::Created { id, name } => Project {
+				id: *id,
+				name: name.to_owned(),
+				..Default::default()
+			},
+			ProjectEvent::LeaderAssigned { id, leader_id } => {
+				self.leaders.insert(*leader_id);
+				Project { id: *id, ..self }
```
ofux, line 43: Why not returning `self` directly? This is confusing, it looks like the `id` could somehow be changed (even though it's not the case)

ofux (approved):

domain/src/project/aggregate.rs
Bernardstanislas: oh you're right, updating it

api/src/application/project/assign_leader.rs
```diff
@@ -0,0 +1,49 @@
...
+		let events: Vec<Event> = project
+			.assign_leader(leader_id)?
+			.into_iter()
+			.map(|event| Event {
+				deduplication_id: self.uuid_generator.new_uuid().to_string(),
+				event: event.into(),
+				timestamp: Utc::now().naive_utc(),
+				origin: EventOrigin::BACKEND,
+				metadata: Default::default(),
+			})
```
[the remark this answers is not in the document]
ofux: yeah, you're probably right indeed :)

PR #345, opened 2022-11-21
Install Cypress

Bernardstanislas:
Add a very basic test to call the GraphQL endpoint and see if it answers.

Also speed up the CI by running everything in parallel since there is no actual dependencies between jobs **from 16m to 6m**.

TODO: run the Rust backend and test a mutation on it, splitting the PR since it's taking me too much time


.github/actions/cypress/action.yml
```diff
@@ -0,0 +1,23 @@
...
+      shell: bash
+      run: yarn
+      working-directory: ./testing
+
+    - name: Cypress run
+      uses: cypress-io/github-action@v4
+      with:
+        working-directory: ./testing
+      env:
+        CYPRESS_BASE_URL: http://localhost:8080
```
ofux, line 23: Looks like this line is useless since cypress uses dotenv, isn't it?

ofux (approved):

.github/actions/cypress/action.yml
Bernardstanislas: Sure but at this point the `.env` file does not exist

PR #350, opened 2022-11-22
Feature/mutation payment request

ofux:


domain/src/payment_request/aggregate.rs
```diff
@@ -2,19 +2,34 @@ use derive_getters::{Dissolve, Getters};
...
 }
 
 impl Aggregate for PaymentRequest {
-	type Event = ProjectEvent;
+	type Event = PaymentRequestEvent;
 	type Id = PaymentRequestId;
 }
+
+impl PaymentRequest {
+	pub fn create(
```
Bernardstanislas, line 18: Can you test it? Seems useless but we must have full coverage on domain

Bernardstanislas (approved):

AnthonyBuisset (approved):

domain/src/payment_request/aggregate.rs
ofux: will be tested in the next PR (the one adding `specifications`) as there will actually be something to test ;)

PR #353, opened 2022-11-22
feature/specifications

ofux:
- :sparkles: introduce specifications
- :recycle: make aggregates testable again by introducing the Specifications trait
- ✅ add tests for payment request aggregate


domain/src/specifications/mod.rs
```diff
@@ -10,12 +14,17 @@ pub enum Error {
 	EventStore(aggregate_root::Error),
 }
 
-pub struct Specifications {
+#[async_trait]
+pub trait Specifications: ProjectSpecifications + Send + Sync {}
```
AnthonyBuisset, line 18: shouldn't this be the other way around ?
```
pub trait ProjectSpecifications: Specifications {}
```

AnthonyBuisset: I would even imagine having a trait Specifications, implemented by the ProjectSpecification
And then a composite to decide which specification to apply on a given command

domain/src/payment_request/aggregate.rs
```diff
@@ -29,7 +28,7 @@ impl Aggregate for PaymentRequest {
 
 impl PaymentRequest {
 	pub async fn create(
-		specifications: Arc<dyn Specifications>,
+		specifications: &ProjectExistsSpecification,
```
Bernardstanislas, line 31: I think this argument should be called `project_exists_specification`

domain/src/specifications/project_exists.rs
```diff
@@ -1,17 +1,21 @@
...
-#[async_trait]
-impl Specification for SpecificationsImpl {
-	async fn project_exists(&self, project_id: &ProjectId) -> Result<bool, Error> {
+#[automock]
+impl Specification {
+	pub fn new(project_repository: Arc<AggregateRootRepository<Project>>) -> Self {
+		Self { project_repository }
+	}
+
+	pub async fn project_exists(&self, project_id: &ProjectId) -> Result<bool, Error> {
```
Bernardstanislas, line 18: It would be more idiomatic to have this method call `satisfied_by` or `is_satisfied_by`, so that in the aggregate we call `project_exists_specification.is_satisfied_by(project_id)`

domain/src/specifications/user_exists.rs
```diff
@@ -0,0 +1,23 @@
+use crate::{
+	specifications::{Error, SpecificationsImpl},
+	AggregateRootRepositoryError, ProjectId,
+};
+use async_trait::async_trait;
+
+#[async_trait]
+pub trait Specification {
```
Bernardstanislas, line 8: I'm not sure it's used yet?

domain/src/specifications/project_exists.rs
```diff
@@ -0,0 +1,27 @@
...
+	project_repository: Arc<AggregateRootRepository<Project>>,
+}
+
+#[automock]
+impl Specification {
+	pub fn new(project_repository: Arc<AggregateRootRepository<Project>>) -> Self {
+		Self { project_repository }
+	}
+
+	pub async fn is_satisfied_by(&self, project_id: &ProjectId) -> Result<bool, Error> {
```
AnthonyBuisset, line 18: no need to be async for now I guess

AnthonyBuisset (approved):

PR #358, opened 2022-11-23
prevent payment creation for non existing payment request

AnthonyBuisset:
- 🚚 rename specification to make it generic
- ♻️ make aggregate exists spec generic
- ✅ add specification unit test
- ✨ create payment request exists specification
- ✨ payment request is an aggregate root
- ♻️ remove useless Arc
- ✨ prevent payment creation for non existing payment request


Bernardstanislas (approved):
Yeeaahhh

PR #361, opened 2022-11-24
implement user repository with a graphql client targeting Hasura

ofux:
- ✨ add command to export Hasura schema
- ✨ implement user repository with a graphql client


infrastructure/src/graphql/hasura_client.rs
```diff
@@ -0,0 +1,25 @@
...
+
+fn graphql_url() -> String {
+	std::env::var("HASURA_GRAPHQL_GRAPHQL_URL").expect("HASURA_GRAPHQL_GRAPHQL_URL must be set")
+}
+
+fn hasura_secret() -> String {
+	std::env::var("HASURA_GRAPHQL_ADMIN_SECRET").expect("HASURA_GRAPHQL_ADMIN_SECRET must be set")
+}
+
+pub fn new() -> GraphQLClient {
```
Bernardstanislas, line 13: It feels weird to me that this is a function approach, whereas the `GraphQLClient` is a struct. We could have imagined a `HasuraClient` struct too

infrastructure/src/graphql/hasura_client.rs
```diff
@@ -0,0 +1,25 @@
...
+	let mut custom_headers = HeaderMap::new();
+	custom_headers.insert(
+		"x-hasura-admin-secret",
+		HeaderValue::from_str(hasura_secret().as_str())
+			.expect("HASURA_GRAPHQL_ADMIN_SECRET has an invalid value"),
+	);
+	GraphQLClient {
```
Bernardstanislas, line 20: You might want to use `GraphQLClient::new` right?

Bernardstanislas (approved):

infrastructure/src/graphql/hasura_client.rs
AnthonyBuisset: As `GraphQLClient` is a struct we own, I am ok with this approach. But I would propose some renaming:
```
mod hasura {

fn admin_secret() -> String {
}

fn create_client() -> GraphQLClient {
}

}
```
I would keep the `new()` keyword for public methods (i.e. HasuraClient::new())

IMO, creating a `HasuraClient` struct is a bit overkill as it would require re-defining all methods of `GraphQLClient` whereas only the construction part changes (hence the factory function)

AnthonyBuisset (approved):

infrastructure/src/graphql/hasura_client.rs
ofux: I actually changed and made it the other way around. It feels better to have a struct HasuraClient that implements the UserRepository (rather than a generic GraphQLClient).
Instead, the GraphQLClient just becomes a high-level client that is used internally by HasuraClient. Please, check my last commit :)

infrastructure/src/graphql/hasura_client.rs
ofux: definitely, this was the intention

Bernardstanislas (approved):

infrastructure/src/graphql/hasura_client.rs
AnthonyBuisset: Clean !

AnthonyBuisset (approved):

PR #362, opened 2022-11-24
merge payment request and receipt in same aggregate

AnthonyBuisset:
- 🔥 payment request is no longer event sourcable
- ♻️ move payment_request::create into payment::request
- ♻️ move payment::create to payment::process
- ⚰️ remove legacy payment_request aggregate


domain/src/payment/aggregate.rs
```diff
@@ -17,33 +19,82 @@ impl Aggregate for Payment {
...
+	RecipientNotFound,
 	#[error(transparent)]
 	Specification(specifications::Error),
 }
 
 impl Payment {
-	pub fn create(
-		payment_request_exists: &PaymentRequestExists,
+	#[cfg_attr(feature = "cargo-clippy", allow(clippy::too_many_arguments))]
+	pub async fn request(
```
ofux, line 47: I think we should also check that the payment doesn't already exist here, as a payment cannot be requested it is was already processed for example.

domain/src/payment/aggregate.rs
```diff
@@ -17,33 +19,82 @@ impl Aggregate for Payment {
...
+			id,
+			project_id,
+			requestor_id,
+			recipient_id,
+			amount_in_usd,
+			reason,
+		}])
+	}
+
+	pub fn add_receipt(
```
ofux, line 88: Check the payment exists (ie. was requested)?

ofux (approved):
LGTM, we just need a few more checks I think

api/src/application/payment/request.rs
```diff
@@ -32,20 +32,20 @@ impl Usecase {
...
 		requestor_id: UserId,
 		recipient_id: UserId,
 		amount_in_usd: u32,
 		reason: Value,
 	) -> Result<PaymentId> {
-		let payment_request_id = self.uuid_generator.new_uuid();
+		let payment_id = self.uuid_generator.new_uuid();
 
-		PaymentRequest::create(
+		Payment::request(
```
Bernardstanislas, line 45: Crystal clear

domain/src/payment/aggregate.rs
Bernardstanislas: Cannot be called if not event-sourced, so this check is already done

domain/src/payment/aggregate.rs
```diff
@@ -85,16 +85,16 @@ impl Payment {
 		}])
 	}
 
-	pub fn create(
+	pub fn add_receipt(
 		id: PaymentId,
```
Bernardstanislas, line 89: You don't need the id since it's the aggregate you are calling the method on

domain/src/payment/aggregate.rs
```diff
@@ -23,7 +23,7 @@ impl EventSourcable for Payment {
 	fn apply_event(self, event: &Self::Event) -> Self {
 		match event {
 			PaymentEvent::Requested { id, .. } => Self { id: *id },
-			PaymentEvent::Created { id, .. } => Self { id: *id },
+			PaymentEvent::Processed { id, .. } => Self { id: *id },
```
Bernardstanislas, line 26: This should not create an aggregate instance, it should push in a private field a new payment receipt

domain/src/payment/aggregate.rs
Bernardstanislas: You should have `&self` though

domain/src/payment/aggregate.rs
```diff
@@ -85,16 +85,16 @@ impl Payment {
...
-	pub fn create(
+	pub fn add_receipt(
 		id: PaymentId,
-		request_id: PaymentRequestId,
+		receipt_id: PaymentReceiptId,
 		amount: Amount,
 		receipt: PaymentReceipt,
 	) -> Result<Vec<<Self as Aggregate>::Event>, Error> {
-		let events = vec![PaymentEvent::Created {
+		let events = vec![PaymentEvent::Processed {
```
Bernardstanislas, line 94: You should do the monetary checks here, see if the receipt does not exceed the initial amount requested + the already received receipt amounts

Bernardstanislas (changes requested):
I have the feeling you missed the part where you have to store the receipts inside the payment aggregate root? And do the business checks on it

api/src/application/payment/process.rs
```diff
@@ -0,0 +1,44 @@
...
+	}
+
+	pub async fn add_receipt(
+		&self,
+		payment_id: PaymentId,
+		amount: Amount,
+		receipt: PaymentReceipt,
+	) -> Result<PaymentReceiptId> {
+		let receipt_id = self.uuid_generator.new_uuid();
+		let events: Vec<_> = Payment::add_receipt(payment_id, receipt_id.into(), amount, receipt)?
```
Bernardstanislas, line 32: Fetch the payment first, then call `add_receipt` on it. This method should not be static

Bernardstanislas (changes requested):

domain/src/payment/aggregate.rs
AnthonyBuisset: Yes indeed, planned for a next PR, I wanted to keep this one iso-functional

domain/src/payment/aggregate.rs
AnthonyBuisset: Yes indeed, but I planned to add it in another PR along with the amount check

api/src/application/payment/process.rs
Bernardstanislas: Seen IRL, you do it in the next PR

AnthonyBuisset: I have updated the PR, new commits include this change

Bernardstanislas: ok I'm reviewing it

domain/src/payment/aggregate.rs
```diff
@@ -5,13 +5,16 @@ use crate::{
...
 };
+use rust_decimal::Decimal;
 use serde::{Deserialize, Serialize};
 use serde_json::Value;
 use thiserror::Error;
 
 #[derive(Debug, Default, Clone, PartialEq, Eq, Serialize, Deserialize)]
 pub struct Payment {
 	id: PaymentId,
+	requested_usd_amount: Decimal,
```
Bernardstanislas, line 16: Oh even simpler than what I had in mind with the vec of payments, great!

domain/src/payment/aggregate.rs
```diff
@@ -86,13 +100,21 @@ impl Payment {
...
 
 	pub fn add_receipt(
-		id: PaymentId,
+		&self,
 		receipt_id: PaymentReceiptId,
 		amount: Amount,
 		receipt: PaymentReceipt,
 	) -> Result<Vec<<Self as Aggregate>::Event>, Error> {
+		// TODO: Handle currency conversion when needed
+		println!("paid_usd_amount: {}", self.paid_usd_amount);
```
Bernardstanislas, line 109: Put some traces maybe?

Bernardstanislas (approved):

domain/src/payment/aggregate.rs
AnthonyBuisset: indeed ! leftover of debugging, will add a nice trace instead 👍

PR #363, opened 2022-11-25
add some e2e tests

AnthonyBuisset:
- ✅ make e2e tests independent
- ✅ refactor e2e tests to use cypress commands
- ✅ add cypress command to create a user
- ✅ test assign leader to project

### DIFF
--- a/testing/cypress/e2e/retool.cy.js
+++ b/testing/cypress/e2e/retool.cy.js
@@ -2,23 +2,7 @@ describe("Retool", () => {
     it("can create projects", () => {
         const projectName = "Cypress test project";
 
-        cy.request({
-            method: "POST",
-            url: "/v1/graphql",
-            body: {
-                query: `
-                    mutation{
-                        createProject(
-                            name: "${projectName}"
-                        )
-                    }`,
-            },
-            headers: {
-                "X-Hasura-Admin-Secret": Cypress.env("hasuraAdminSecret"),
-            },
-        })
-            .its("body.data.createProject")
-            .should("be.a", "string")
+        cy.createProject(projectName)
             .then($project_id => {
                 // Let the event sourcing magic happen
                 cy.wait(500);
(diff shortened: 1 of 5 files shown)


testing/cypress/e2e/retool.cy.js
```diff
@@ -2,35 +2,43 @@ describe("Retool", () => {
...
+                    .its("body")
+                    .should("deep.equal", {
+                        data: {
+                            projects: [{ name: projectName }],
+                        },
+                    });
             });
     });
+
+    it('can assign a leader to a project', () => {
```
lechinoix, line 22: You can probably use [async/await syntax](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/async_function) to avoid callback hell ^^

testing/cypress/support/commands.js
```diff
@@ -7,19 +7,66 @@
...
+
+    cy.request({
+        method: 'POST',
+        url: 'http://localhost:4000/signup/email-password',
+        body: {
+            "email": email,
+            "options": {
+                "allowedRoles": [
+                    "me",
+                    "public"
```
lechinoix, line 47: You are missing a "user" role that I used in ACLs (actually I do not use "me", but I assumed it might be a role for hasura auth internals)

testing/cypress/e2e/retool.cy.js
AnthonyBuisset: Actually I can't :-(
CleanShot 2022-11-25 at 15 35 34

lechinoix: Ah nooooo

lechinoix (approved):

testing/cypress/support/commands.js
AnthonyBuisset: indeed, fixed

PR #368, opened 2022-11-29
feature/project details

ofux:
- 🗃️ create project_details table
- ✨ add project_details entity and repository
- ✨ update create project mutation to add project_details
- ✅ update e2e tests
- ✨ make project_details queryable through Hasura

Notes:
- I decided to stick to a single `github_repo_id` per project for now (as described in the task), as supporting multiple ones needs another refinement.
- I didn't encapsulate the publish+insert inside a single DB transaction because diesel doesn't expose a satisfying API to do so. I decided to simply do the `insert` after the `publish`. This means the `insert` could fail while the `publish` succeeded.
- I chose to use the `project_details_repository` directly from the usecase, because it looked overkill to introduce a pass-through service to do the same thing in the end. I don't have a strong opinion on this, but I feel like it is actually a non-issue.


api/src/infrastructure/database/models/project_details.rs
```diff
@@ -0,0 +1,23 @@
+use crate::domain::ProjectDetails;
+use ::infrastructure::database::schema::*;
+use diesel::{AsChangeset, Insertable};
+use serde::{Deserialize, Serialize};
+
+#[derive(Insertable, Debug, Serialize, Deserialize, AsChangeset)]
+pub struct ProjectDetail {
```
AnthonyBuisset, line 7: why not put this in the domain directly and avoid an additional struct ?

api/src/infrastructure/database/repositories/project_details.rs
```diff
@@ -0,0 +1,33 @@
+use crate::{
+	database::{schema::project_details::dsl, Client},
+	domain::{ProjectDetailsRepository, ProjectDetailsRepositoryError},
+	infrastructure::database::models,
+};
+use diesel::RunQueryDsl;
+use tracing::error;
+
+impl ProjectDetailsRepository for Client {
```
AnthonyBuisset, line 9: Could be generated using the macro DeriveDieselRepository ?

AnthonyBuisset:
would be nice to add an e2e test to check that values are actually stored and retrievable using hasura graphQL

AnthonyBuisset (approved):

PR #369, opened 2022-11-29
Spend budget when requesting payment

AnthonyBuisset:
* Create budget aggregate root and projection
* Check remaining amount when spending budget
* Spend budget when requesting a payment
* The payment requets projection is now linked to a budget (instead of a project). The budget is linked to a project.
* Removed `project_exists` specification as checked by the budget event sourcing logic

I had to remove the access control on `Payment::request` as we need to refine how a budget can be linked to a user (owner?)

Commits:
- 🚚 move amount as value object
- ✨ create budget aggregate
- ✨ allocate budget upon project creation
- 🗃️ create budget projection
- ♻️ use derive_new to remove boilerplate
- ✨ create budget projection
- 🔧 track budgets table in hasura
- 🗃️ make request_id mandatory in payments projection
- ✨ implement find_by_id in diesel repository
- ✨ event source budget and update projection on spend
- 🗃️ link payment request projection to budget instead of project
- 🔥 remove project exists spec
- 🔥 remove access control check when requesting payment
- ✨ spend budget when requesting payment

### DIFF
--- a/event-listeners/src/domain/projections/budget.rs
+++ b/event-listeners/src/domain/projections/budget.rs
@@ -1,0 +1,16 @@
+use super::Projection;
+use infrastructure::database::schema::budgets;
+use rust_decimal::Decimal;
+use uuid::Uuid;
+
+#[derive(Debug, Insertable, Identifiable, AsChangeset, new)]
+pub struct Budget {
+	id: Uuid,
+	project_id: Uuid,
+	initial_amount: Decimal,
+	remaining_amount: Decimal,
+}
+
+impl Projection for Budget {
+	type Id = Uuid;
+}
--- a/event-listeners/src/domain/projections/budget.rs
+++ b/event-listeners/src/domain/projections/budget.rs
@@ -9,7 +9,7 @@ pub struct Budget {
 	id: Uuid,
 	project_id: Option<Uuid>,
 	initial_amount: Decimal,
-	remaining_amount: Decimal,
+	pub remaining_amount: Decimal,
 }
 
 impl Entity for Budget {
--- a/event-listeners/src/domain/projections/budget.rs
+++ b/event-listeners/src/domain/projections/budget.rs
@@ -1,16 +1,19 @@
 use super::Projection;
+use domain::Entity;
 use infrastructure::database::schema::budgets;
 use rust_decimal::Decimal;
 use uuid::Uuid;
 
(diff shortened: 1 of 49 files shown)


event-listeners/src/domain/projections/budget.rs
```diff
@@ -0,0 +1,16 @@
+use super::Projection;
+use infrastructure::database::schema::budgets;
+use rust_decimal::Decimal;
+use uuid::Uuid;
+
+#[derive(Debug, Insertable, Identifiable, AsChangeset, new)]
+pub struct Budget {
+	id: Uuid,
+	project_id: Uuid,
```
ofux: I think this should either be an `Option<Uuid>`, or the SQL column should not be nullable.

migrations/2022-11-28-161958_add-budget-in-project-projection/up.sql
```diff
@@ -0,0 +1,19 @@
+CREATE TABLE budgets (
+    id UUID PRIMARY KEY DEFAULT gen_random_uuid(),
+    project_id UUID REFERENCES projects(id),
```
ofux, line 3: `REFERENCES` => are we sure the project will be projected before the budget?

migrations/2022-11-29-094546_link-payment-to-budget/down.sql
```diff
@@ -0,0 +1,21 @@
+ALTER TABLE payment_requests RENAME TO payment_requests_backup;
```
ofux, line 1: this file is the same as `up.sql`

ofux (changes requested):
A request change as `down.sql` seems to be the same as `up.sql`

migrations/2022-11-29-094546_link-payment-to-budget/down.sql
AnthonyBuisset: it looks like it but it is not :-)
The `up.sql` replaces `project_id` by `budget_id`
The `down.sql` does the opposite

event-listeners/src/domain/projections/budget.rs
AnthonyBuisset: The column is indeed nullable, but here I want to insert a budget that is linked to a project. There is no `None` case for now

PR #372, opened 2022-12-01
macro/NewtypeToSql

ofux:
- ♻️ introduce NewtypeToSql macro
- ♻️ replace ToSql implementation with NewtypeToSql macro

### DIFF
--- a/macros/src/to_sql.rs
+++ b/macros/src/to_sql.rs
@@ -59,7 +59,7 @@ fn get_inner_type(ast: &DeriveInput) -> TypePath {
 	let inner_field = match ast.data {
 		syn::Data::Struct(ref data) => match data.fields {
 			syn::Fields::Unnamed(ref fields) => {
-				assert!(fields.unnamed.clone().len() == 1);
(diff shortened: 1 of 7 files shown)


macros/src/to_sql.rs
```diff
@@ -0,0 +1,97 @@
...
+	sql_type.parse().expect("Invalid Rust type")
+}
+
+fn get_inner_type(ast: &DeriveInput) -> TypePath {
+	const NEWTYPE_ERROR: &str = "Excpected struct with a single unnamed field (newtype pattern)";
+
+	let inner_field = match ast.data {
+		syn::Data::Struct(ref data) => match data.fields {
+			syn::Fields::Unnamed(ref fields) => {
+				assert!(fields.unnamed.clone().len() == 1);
```
AnthonyBuisset, line 62: you can add the NEWTYPE_ERROR as message in the assert:
```
assert_eq!(fields.unnamed.clone().len(), 1, NEWTYPE_ERROR);
```

AnthonyBuisset (approved):

macros/src/to_sql.rs
ofux: good catch. done

PR #375, opened 2022-12-01
Speed up ci

Bernardstanislas:
Kaizen can be seen here https://www.figma.com/file/UJ5rrYIcTjhNXxd44JefPk/Kaizen-CI?node-id=0%3A1&t=ox2LxwFILJuDZn22-1
Still a lot to do, but we're getting close!

Before/after:
<div style="display:flex">
<div>

</div>

N.B.: performance improvement is widely due to cache fine tuning, so whenever cache is missed for any given reason, CI time triples

### DIFF
--- a/.github/actions/coverage-checks/action.yml
+++ b/.github/actions/coverage-checks/action.yml
@@ -5,10 +5,16 @@ description: Launch application and run all tests while measuring code coverage
 runs:
   using: "composite"
   steps:
+    - uses: Swatinem/rust-cache@v2
+      with:
+        workspaces: ". -> tarpaulin_target"
+
     - name: Run cargo-tarpaulin
       uses: actions-rs/tarpaulin@v0.1
       with:
         version: "0.20.1"
         args: |
           --workspace
+          --skip-clean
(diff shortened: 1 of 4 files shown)


.github/actions/coverage-checks/action.yml
```diff
@@ -5,10 +5,16 @@ description: Launch application and run all tests while measuring code coverage
...
+      with:
+        workspaces: ". -> tarpaulin_target"
+
     - name: Run cargo-tarpaulin
       uses: actions-rs/tarpaulin@v0.1
       with:
         version: "0.20.1"
         args: |
           --workspace
+          --skip-clean
```
AnthonyBuisset, line 18: I had to remove this in the past as otherwise it was crashing

AnthonyBuisset (approved):

.github/actions/coverage-checks/action.yml
Bernardstanislas: Ok, with the current method we have a finer control over cargo builds so I guess this is why it is working now

PR #378, opened 2022-12-02
create github proxy backend with simple graphql API

AnthonyBuisset:
- 🎉 initiate github-proxy back-end
- ✨ create web server using rocket
- ✨ expose graphql endpoint
- 🔧 add github proxy as hasura remote schema
- ✅ add e2e connectivity test
- 👷 add github-proxy in CI for e2e tests


github-proxy/src/presentation/http/mod.rs
```diff
@@ -1,8 +1,19 @@
 use anyhow::Result;
 
 mod config;
+mod routes;
 
 pub async fn serve() -> Result<()> {
-	let _ = rocket::custom(config::get()).launch().await?;
+	let _ = rocket::custom(config::get())
```
Bernardstanislas, line 7: Why not presenting it on the backend API? They don't have the same role, but that's the essence of an API, it's ok to expose several different services under a given endpoint.
Having two APIs implies having more infrastructure setup and costs

Procfile
```diff
@@ -1,4 +1,5 @@
 web: AMQP_ADDR=$CLOUDAMQP_URL ROCKET_PORT=$PORT ./target/release/api
 event-store: AMQP_ADDR=$CLOUDAMQP_URL ./target/release/event-store
 event-listeners: AMQP_ADDR=$CLOUDAMQP_URL WEBSERVER_PORT=$PORT ./target/release/event-listeners
+github-proxy: ROCKET_PORT=$PORT ./target/release/github-proxy
```
Bernardstanislas, line 4: this won't work, only one app can bind to a port, and it has to be the `web` dyno in Heroku

AnthonyBuisset: As discussed altogether we will create a new APP on heroku for this new BE

github-proxy/src/presentation/http/mod.rs
AnthonyBuisset: Because they have different purposes, so best to segregate them.
Also, as discussed together, the infra maintenance cost is not very high

PR #380, opened 2022-12-05
♻️ refactor trace_id conversion to do the exact same thing as opentelemetry_datadog crate

ofux:


olog/src/lib.rs
```diff
@@ -11,18 +11,16 @@ pub use tracing_opentelemetry;
...
 				),
 			)
 			.span_context()
 			.trace_id()
-			.to_bytes()[..8]
-				.try_into()
-				.unwrap_or_default(),
-		)
+			.to_bytes(),
+		) as u64
```
AnthonyBuisset, line 23: `u128` as `u64`, is it safe ? what does it change ?

PR #384, opened 2022-12-05
Move app configuration to toml file and split .env by backend

AnthonyBuisset:
- 🎉 initiate github-proxy back-end
- ✨ create web server using rocket
- ✨ expose graphql endpoint
- 🔧 add github proxy as hasura remote schema
- ✅ add e2e connectivity test
- 👷 add github-proxy in CI for e2e tests
- 🔧 declare dyno in Procfile
- 🚚 move Rocket.toml in api crate
- 🔧 remove useless env variables
- 🔧 move DATABASE_URL from .env to config file
- 🔧 move AMQP_ADDR env variable to config file
- 🔧 move HASURA_GRAPHQL_GRAPHQL_URL to config file
- 🔧 move HASURA_GRAPHQL_ADMIN_SECRET to config file
- 🔧 move CYPRESS_BASE_URL to config file
- 🔧 move NASI_LOGS to config file
- 🔧 move .env api config into its own crate


.env.example
```diff
@@ -1,9 +1,6 @@
 # API
 BACKEND_GRAPHQL_API_KEY=myapikey
 
-# Database
-DATABASE_URL=postgres://postgres:postgres@localhost/marketplace_db
```
Bernardstanislas, line 5: How do I override this with an env var now? Something like `ROCKET_DATABASE_URL`?

.env.example
```diff
@@ -7,26 +7,10 @@ DATABASE_URL=postgres://postgres:postgres@localhost/marketplace_db
...
 # Logs
 ANSI_LOGS=true
 RUST_LOG=info
 
-# Heroku
-HEROKU_TOKEN="1234"
-HEROKU_APP_NAME="onlydust-backend-staging"
-
-# Hasura Auth
-AUTH_PROVIDER_GITHUB_CLIENT_ID=my_github_client_id
```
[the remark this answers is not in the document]
AnthonyBuisset: they are already in the docker, so not mandatory in the .env. Can still be overridden though

.env.example
AnthonyBuisset: you can still use `DATABASE_URL` as the mapping from env is done in the code

AnthonyBuisset:
After discussing this PR with @ofux  I won't be going forward with it. It is too soon.
Most of the configuration are local only.
Secrets should be kept in the `.env`, and as `DATABASE_URL` contains the credentials, it should be hidden as well.
Other variables like the `BACKEND_GRAPHQL_API_KEY` is shared between Hasura and the BE, so cannot be renamed.
The outcome of this refactoring feels weird, it is a sign it is not needed for now.
Let's keep it in mind when we will actually need it.

PR #390, opened 2022-12-07
create application configuration

AnthonyBuisset:
It is not perfect (especially the infrastructure tests), but it's a start.
Let's build on top of it.


app.json
```diff
@@ -15,10 +15,8 @@
     ],
     "env": {
         "PG_POOL_MAX_SIZE": "3",
-        "ROCKET_ADDRESS": "0.0.0.0",
```
Bernardstanislas, line 18: hihi

app.json
```diff
@@ -15,10 +15,8 @@
...
         "ROCKET_CLI_COLORS": "false",
-        "RUST_LOG": "debug",
-        "SLOG_CHANNEL_SIZE": "1024"
```
Bernardstanislas, line 21: Did you remove these variables directly in the Heroku apps too?

infrastructure/src/config.rs
```diff
@@ -0,0 +1,42 @@
...
+	use super::*;
+
+	#[derive(Deserialize)]
+	pub struct App {
+		pub api_key: String,
+	}
+
+	#[test]
+	fn config_from_file() {
+		figment::Jail::expect_with(|jail| {
```
Bernardstanislas: 🤯

api/src/app.yaml
```diff
@@ -0,0 +1,14 @@
+local:
+  database:
+    url: postgres://postgres:postgres@localhost/marketplace_db
+    pool_max_size: 20
+
+staging:
```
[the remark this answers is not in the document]
Bernardstanislas: I mean it's a good sign to see "production" profile in a staging app config for example

api/src/app.yaml
```diff
@@ -4,17 +4,26 @@ local:
...
+    admin_secret: $HASURA_GRAPHQL_ADMIN_SECRET
 
 staging:
   database:
     url: $DATABASE_URL
     pool_max_size: 3
   amqp:
     url: $CLOUDAMQP_URL
+  graphql:
+    url: https://staging.hasura.onlydust.xyz/v1/graphql
```
Bernardstanislas: Here for instance, one could argue that the distinction between staging and production is necessary to make them point at different endpoints. I would say that this means this endpoint should live in an env var, as stated by https://12factor.net/config

infrastructure/src/database/config.rs
```diff
@@ -1,7 +1,8 @@
 use derive_getters::Getters;
+use derive_more::Constructor;
 use serde::Deserialize;
 
-#[derive(Deserialize, Getters)]
+#[derive(Deserialize, Getters, Constructor)]
```
Bernardstanislas, line 5: For my own culture, what does it do?

app.json
AnthonyBuisset: not yet, I will cleanup once code is loaded

api/src/app.yaml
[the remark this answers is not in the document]
AnthonyBuisset: there could be different settings, for instance the graphql server URL is different

infrastructure/src/database/config.rs
AnthonyBuisset: creates the function `new` for you ❤️

api/src/app.yaml
AnthonyBuisset: fair point, I can change that
this would also allow to put the `PROFILE` var in the `app.json`

Bernardstanislas (approved):

PR #400, opened 2022-12-09
protect github proxy with api_key

AnthonyBuisset:
- 🎉 create presentation crate
- 🚚 move rocket common logic in presentation crate
- 🔒️ protect github proxy by api key
- ✨ add common routes in github proxy


github-proxy/src/presentation/http/mod.rs
```diff
@@ -7,8 +7,16 @@ mod routes;
 
 pub async fn serve(config: http::Config, github_service: Arc<dyn GithubService>) -> Result<()> {
 	let _ = rocket::custom(http::config::rocket("github-proxy/Rocket.toml"))
+		.attach(http::guards::Cors)
```
Bernardstanislas, line 10: Do we still want to have CORS on our apis? From my understanding, the frontend only talks to hasura and hasura auth, the Rust APIs are then accessed from backends, not browsers

Bernardstanislas (approved):

github-proxy/src/presentation/http/mod.rs
AnthonyBuisset: removed

PR #401, opened 2022-12-09
payment request github user

ofux:
- ✨ link payment requests to github user ID instead of user id
- 🔥 remove UserExists specification and UserRepository
- 🛂 update Hasura permissions
- ✅ update e2e tests


domain/src/payment/aggregate.rs
```diff
@@ -60,22 +58,13 @@ pub enum Error {
...
-		user_exists_specification: &UserExists,
 		id: PaymentId,
 		budget_id: BudgetId,
 		requestor_id: UserId,
 		recipient_id: GithubUserId,
 		amount_in_usd: u32,
 		reason: Value,
 	) -> Result<Vec<<Self as Aggregate>::Event>, Error> {
-		if !user_exists_specification
-			.is_satisfied_by(&requestor_id)
```
Bernardstanislas, line 72: Why don't we check this anymore?

Bernardstanislas (approved):
From what I understand, there is no user existence check in the domain since we rely on Hasura to ensure the requestor always exists and is allowed to do the action.
There can be a debate on this, maybe with a very quick ADR to trace this decision?

And beware that once deployed, this PR will break the frontend, might need to sync with @oscarwroche

domain/src/payment/aggregate.rs
ofux: because we have this check in the mutation:
```rust
if !context.caller_permissions.can_spend_budget(&budget_id.into()) {
			return Err(Error::NotAuthorized(
				"Budget spender role required".to_string(),
			));
		}
```
However, I wanted to discuss this: the check that is done in the mutation is the right one, but I think it should be within the domain... which means the Permissions object should be in the domain itself. WDYT?

testing/cypress/e2e/simple_user.cy.js
```diff
@@ -116,56 +116,93 @@ describe("As a simple user, I", () => {
 
     it("can update my info", () => {
         let email = "pierre.fabre@gmail.com";
-        let location = '{city: "Paris", country: "France", number: "4", postCode: "75008", street: "avenue des Champs Elysee"}';
-        let identity = '{type: PERSON, optPerson: {firstname: "Pierre", lastname: "Fabre"}}';
-        let payout_settings = '{type: ETHEREUM_ADDRESS, optEthAddress: "0x123"}';
+        let location =
```
tdelabro, line 119: Which linter do you use?

api/app.yaml
```diff
@@ -8,9 +8,6 @@ local:
     pool_max_size: 20
   amqp:
     url: "amqp://127.0.0.1:5672/%2f"
-  graphql:
```
AnthonyBuisset, line 11: can you please clean-up heroku as well ?

api/src/presentation/graphql/mutation.rs
```diff
@@ -129,7 +129,7 @@ impl Mutation {
 			.request(
 				budget_id.into(),
 				caller_id,
-				recipient_id.into(),
+				(recipient_id as i64).into(),
```
tdelabro, line 132: why is one i64 and the other i32?
Moreover do we really think some id will be negative?

AnthonyBuisset: those are limitations from crates
`junyper` only supports `i32`, `octocrab` requires a `i64` but actually, ids are never negative...

AnthonyBuisset (approved):

api/app.yaml
ofux: you mean env vars in Heroku?

testing/cypress/e2e/simple_user.cy.js
ofux: prettier. We actually need to add a conf so we all have the same formatting

api/src/presentation/graphql/mutation.rs
ofux: indeed

PR #421, opened 2022-12-15
feat: remove the need to impl Entity when Identifiable is derived

tdelabro:


backend/common/domain/src/entity.rs
```diff
@@ -1,5 +1,15 @@
...
 	type Id: PartialEq + Display;
 }
+
+impl<'ident, T> Entity for &'ident T
+where
+	&'ident T: Identifiable,
+	<&'ident T as Identifiable>::Id: Display,
+{
+	type Id = <Self as Identifiable>::Id;
+}
```
ofux, line 15: Great idea!
By curiosity, why did you have to pass the lifetime? Wouldn't this work identically if you removed 'ident?

ofux (approved):

backend/common/domain/src/entity.rs
tdelabro: When you derive identifiable it's not derived on `T`. It's derived on `&'ident T` instead. 
So I did the same here.

tdelabro: And you cannot use a ref without a lifetime in a trait implem

PR #426, opened 2022-12-16
fix(EventListener): rework error management to prenvent crash if inse…

tdelabro:
…rtion in db failed

### DIFF
--- a/backend/api/src/application/project/create.rs
+++ b/backend/api/src/application/project/create.rs
@@ -7,6 +7,7 @@ use domain::{
 };
 
 use crate::{
+	application::UsecaseError,
 	domain::{ProjectDetails, Publishable},
 	infrastructure::database::ProjectDetailsRepository,
 };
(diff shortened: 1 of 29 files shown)


ofux (approved):

backend/api/src/application/usecase_error.rs
```diff
@@ -0,0 +1,38 @@
+use domain::{AggregateRootRepositoryError, PublisherError};
+use infrastructure::database::DatabaseError;
```
Bernardstanislas, line 2: The application layer should not know about the infrastructure, can you put the `impl From<DatabaseError> for UsecaseError` in the infra directly?

backend/api/src/application/project/create.rs
```diff
@@ -85,7 +87,7 @@ fn allocate_owned_budget(
 	project_id: ProjectId,
 	owner_id: UserId,
 	initial_budget: Amount,
-) -> Result<Vec<Event>> {
+) -> Vec<Event> {
```
Bernardstanislas, line 90: It seems counter intuitive that a domain action cannot fail, this will change when we add business checks in these actions. But since we don't have some so far, ok for this.

Bernardstanislas (commented):
Be careful you broke the layered architecture, we do have some exceptions but here it is not justified

PR #428, opened 2022-12-16
refacto: move From<DatabaseError> for usecaseError implem to domain

tdelabro:


backend/api/src/application/payment/process.rs
```diff
@@ -3,12 +3,10 @@ use std::sync::Arc;
 use anyhow::Result;
 use domain::{
 	AggregateRootRepository, Amount, Destination, Event, Payment, PaymentId, PaymentReceipt,
-	PaymentReceiptId, Publisher, UniqueMessage, UuidGenerator,
+	PaymentReceiptId, Publisher, UniqueMessage, UsecaseError, UuidGenerator,
```
Bernardstanislas, line 6: It is no longer a `UsecaseError`, what do you think about `DomainError`?

tdelabro: It is not currently used, so yeah can be the name

Bernardstanislas (approved):

PR #449, opened 2022-12-22
e 46 make sure githubrepositoryid exists

ofux:
- 🐛 check github repo exists before emitting project created event
- ✅ add integration test
- 🐛 check github repo exists before emitting GithubRepositoryUpdated event

Solves E-46

### DIFF
--- a/backend/api/src/application/project/create.rs
+++ b/backend/api/src/application/project/create.rs
@@ -5,6 +5,7 @@ use domain::{
 	Amount, Budget, BudgetId, DomainError, Event, EventSourcable, GithubRepositoryId, Project,
 	ProjectId, Publisher, UniqueMessage, UserId, UuidGenerator,
 };
+use infrastructure::github;
 
 use crate::{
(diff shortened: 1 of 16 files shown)


backend/api/src/application/project/create.rs
```diff
@@ -15,18 +16,21 @@ pub struct Usecase {
 	uuid_generator: Arc<dyn UuidGenerator>,
 	event_publisher: Arc<dyn Publisher<UniqueMessage<Event>>>,
 	project_details_repository: ProjectDetailsRepository,
+	github: Arc<github::Client>,
```
AnthonyBuisset, line 19: why not injecting here a `Arc<dyn GithubRepoExists>` ?

AnthonyBuisset (approved):

backend/api/src/application/project/create.rs
ofux: That's a mistake, good catch!

AnthonyBuisset (approved):

PR #479, opened 2022-12-30
validate ENS upon profile update

AnthonyBuisset:
closes E-122

Some difficulties I faced:
 * I could not create a macro to generate the `DieselRepository` mock
 * mocking structs from another crate (like `ens::Client`) is more difficult than mocking a trait, but I did not want to add the overhead of adding a trait for this
 * I did not manage to mock ETH calls using httpmock as we need to capture variables in input to pass then in the response


backend/common/infrastructure/src/config/providers.rs
```diff
@@ -52,11 +53,17 @@ impl TryFromEnv for Vec<Value> {
...
 
 impl TryFromEnv for String {
-	fn try_from_env(self) -> Option<Self> {
-		if let Some(variable) = self.strip_prefix('$') {
-			std::env::var(variable).ok()
-		} else {
-			Some(self)
+	fn try_from_env(mut self) -> Option<Self> {
+		lazy_static! {
+			static ref RE: Regex = Regex::new(r"(?:\$)(?P<var>[[:alpha:]_]+)").unwrap();
```
tdelabro, line 58: Can you add a comment explaining what this regex does?

backend/api/src/domain/specifications/payout_settings.rs
```diff
@@ -0,0 +1,95 @@
...
+#[cfg(not(test))]
+use infrastructure::web3::ens::Client as EnsClient;
+use infrastructure::web3::ens::Error as EnsError;
+#[cfg(test)]
+use tests::MockEnsClient as EnsClient;
+
+use crate::domain::user_info::{EthereumIdentity, PayoutSettings};
+
+#[derive(Constructor)]
+pub struct IsValid {
```
tdelabro, line 14: I feel like this is more EnsClient wrapper.
It may serve other purposes, rather than just checking that a name exists.

I would rewrite this part with and `EnsClient` struct and a `does_domain_exist` method.

tdelabro (approved):

backend/common/infrastructure/src/web3/ens.rs
```diff
@@ -0,0 +1,36 @@
...
+use web3::{api::Namespace, contract::ens::Ens, transports::Http};
+
+use super::Config;
+
+#[derive(Debug, Error)]
+pub enum Error {
+	#[error("Provider URL is not valid")]
+	InvalidProviderUrl(#[source] anyhow::Error),
+	#[error("Unable to call ENS resolver")]
+	Contract(#[source] anyhow::Error),
```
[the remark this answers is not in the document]
AnthonyBuisset: This is not to leak the dependency to other crates. web3 is an implementation detail.
I started without it and I had to add the web3 dependency to other crates using it just for error management

backend/common/infrastructure/src/config/providers.rs
AnthonyBuisset: sure

PR #480, opened 2023-01-02
Unassign budget lead

tdelabro:


backend/api/src/presentation/graphql/mutation.rs
```diff
@@ -154,6 +156,28 @@ impl Mutation {
...
+				"Only an admin can unasign an project lead".to_string(),
+			));
+		}
+
+		context
+			.remove_project_leader_usecase
+			.remove_leader(&project_id, &user_id)
+			.await?;
+
+		Ok(1)
```
tdelabro, line 179: We have no uuid to return and `()` is not a valid scalar type, so I opted to return something. 
It can also be a bool if we want. wdyt

backend/common/domain/src/project/aggregate.rs
```diff
@@ -10,6 +10,8 @@ use crate::*;
 pub enum Error {
 	#[error("Project lead already assigned to this project")]
 	LeaderAlreadyAssigned,
+	#[error("User is not a project leader")]
```
ofux, line 13: Shouldn't we add more context to errors when we can, like, for example here the user id. I think this kind of information will be very valuable if we need to debug production issues.

This comment also applies to other errors (like the one above) and is not specific to this PR.

ofux (approved):

backend/common/domain/src/project/aggregate.rs
tdelabro: The assertion is that User, is the user making the call. Do we have knowledge of who is making the call in the logs ?

tdelabro: But yeah, at worst it will be duplicated info in two places. Still valuable

backend/api/src/presentation/graphql/mutation.rs
Bernardstanislas: let's go for bool, less cryptic than `1` IMHO

backend/common/domain/src/project/aggregate.rs
ofux: you're right but as you said, it is always harder to debug when we need to correlate data to find information. Duplicated information in logs is not an issue IMO

PR #507, opened 2023-01-11
Adjust sidebar design according to specs

Bernardstanislas:
Closes E-166


AnthonyBuisset:
* Cards height are a bit too big, they make the screen scroll when not needed
* Some project images have the wrong size in the sidebar
* The sidebar width changes when opening it, is it possible to fix it ?

AnthonyBuisset (approved):

PR #518, opened 2023-01-12
E 196 format validate iban bic

oscarwroche:


oscarwroche:
Still need to fix e2e tests but you can review

PR #527, opened 2023-01-14
Fix some QA faults

AnthonyBuisset:
- My projects navigate to the latest visited project (default to first led project)
- When clicking on a project from "Browse project", redirect to "My project" if I lead the project or am invited
- Sidebar dropdown covers the content
- Set cursor pointer when hovering over sortable table header


frontend/src/hooks/useSession.tsx
```diff
@@ -0,0 +1,39 @@
...
+
+  const value = {
+    lastVisitedProjectId,
+    setLastVisitedProjectId,
+  };
+
+  return <SessionContext.Provider value={value}>{children}</SessionContext.Provider>;
+};
+
+export const useSession = (): SessionContextType => {
```
Bernardstanislas, line 33: I think we should unit test all of our hooks

frontend/src/pages/ProjectDetails/index.tsx
```diff
@@ -118,6 +109,12 @@ export default function ProjectDetails({ onlyMine = false }: ProjectDetailsProps
     }
   }, [acceptProjectLeaderInvitationMutation]);
 
+  const isProjectMine = (projectId: string) => {
```
Bernardstanislas: I feel this component starts to have a lot of logic, it is worth splitting it between the component's behaviour logic `index.txs` and the presentation logic `View.tsx`, and then unit test it.
We can alternatively extract the behaviour in a custom, unit-tested, hook, that would be very neat

frontend/src/hooks/useProjectLeadInvitations.tsx
```diff
@@ -0,0 +1,42 @@
+import { useHasuraMutation, useHasuraQuery } from "src/hooks/useHasuraQuery";
+import { HasuraUserRole } from "src/types";
+import { PendingProjectLeaderInvitationsQuery } from "src/__generated/graphql";
+import { useAuth } from "./useAuth";
+import { gql } from "@apollo/client";
+import { ACCEPT_PROJECT_LEADER_INVITATION_MUTATION } from "src/pages/ProjectDetails";
+
+export const useProjectLeadInvitations = () => {
```
Bernardstanislas, line 8: We must unit test this

Bernardstanislas (changes requested):
I would expect all the hooks to be unit-tested, and the logic in the `ProjectDetails` view to be extracted in a custom hook and unit tested.
Also check that the "active" behaviour of the "My projects" link in the header navbar still works, I cannot test it on the review app but I think it can break with those changes.

frontend/src/pages/ProjectDetails/index.tsx
AnthonyBuisset: Indeed, I will do it in a dedicated PR

frontend/src/hooks/useProjectLeadInvitations.tsx
AnthonyBuisset: done

ofux (approved):

Bernardstanislas (approved):

PR #530, opened 2023-01-16
Cancel payment request

ofux:
- ✨ add cancel() method to payment aggregate
- ✨ add cancel_payment_request mutation
- ✅ add e2e tests

### DIFF
--- a/backend/api/src/application/payment/cancel.rs
+++ b/backend/api/src/application/payment/cancel.rs
@@ -1,12 +1,11 @@
 use std::sync::Arc;
 
 use anyhow::Result;
-use domain::{
-	AggregateRootRepository, Destination, DomainError, Event, Payment, PaymentId, Publisher,
-};
-use event_store::bus::QUEUE_NAME as EVENT_STORE_QUEUE;
+use domain::{AggregateRootRepository, DomainError, Event, Payment, PaymentId, Publisher};
 use infrastructure::amqp::UniqueMessage;
 use tracing::instrument;
+
+use crate::domain::Publishable;
 
 pub struct Usecase {
 	event_publisher: Arc<dyn Publisher<UniqueMessage<Event>>>,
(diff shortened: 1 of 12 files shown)


backend/event-listeners/src/domain/projectors/payment_request.rs
```diff
@@ -22,24 +23,31 @@ impl Projector {
...
+						*id,
+						*budget_id,
+						*requestor_id,
+						*recipient_id,
+						*amount_in_usd as i64,
+						reason.clone(),
+					))
+					.map_err(DatabaseError::into),
+				PaymentEvent::Cancelled { id } =>
+					self.repository.delete(id).map_err(DatabaseError::into),
```
AnthonyBuisset, line 47: we should remove the foreign key between `payments` and `payment_requests` as we can submit "partial" payment receipts

backend/api/src/application/payment/cancel.rs
```diff
@@ -0,0 +1,44 @@
...
+		let payment = self.payment_repository.find_by_id(&payment_id)?;
+		let events: Vec<_> = payment
+			.cancel()
+			.map_err(|e| DomainError::InvalidInputs(e.into()))?
+			.into_iter()
+			.map(Event::from)
+			.map(UniqueMessage::new)
+			.collect();
+
+		self.event_publisher
```
AnthonyBuisset, line 38: if you import `Publishable`, you can chain everything with `.publish()` function

AnthonyBuisset (approved):

backend/event-listeners/src/domain/projectors/payment_request.rs
ofux: For now, it is forbidden to cancel a payment as soon as a receipt has been submitted for it (even partial).
I think supporting such things require a proper specification as there are many possibilities with different implications for the user.

AnthonyBuisset: 👍

backend/api/src/application/payment/cancel.rs
ofux: done

PR #556, opened 2023-01-21
Project card list design

Bernardstanislas:
Closes E-240

Did a lot of uniformization work, cards are now very generic. There is still somewhere we lack uniformity with the cards, in the payment estimation card. We need to discuss with the designer to see how we can easily integrate the Figma component.

I also put in common a `Button` component, which we can now use everywhere.

I also played with the stories to have something we can easily use, go and see the project card story to find out :)


frontend/src/components/ProjectInformation/index.tsx
```diff
@@ -42,12 +42,12 @@ export default function ProjectInformation({
...
         <div className="flex flex-row gap-4 items-start">
           <RoundedImage src={details?.logoUrl || onlyDustLogo} alt="Project Logo" size="lg" className="mt-1" />
-          <div className="flex flex-col">
+          <div className="flex flex-col gap-0.5">
             <div className="text-2xl font-medium font-belwe">{name}</div>
             {lead && (
               <div className="text-md text-neutral-300 font-bold flex flex-row gap-1 items-center">
-                <div>Led by </div>
-                <div className="text-purple-700">{lead?.displayName}</div>{" "}
+                <div className="whitespace-nowrap">Led by </div>
```
AnthonyBuisset: should we translate this ?

AnthonyBuisset (approved):

frontend/src/components/ProjectInformation/index.tsx
Bernardstanislas: Yes we should!
Let me add a PR for that

PR #561, opened 2023-01-23
♻️ use github user id from JWT

AnthonyBuisset:


frontend/src/hooks/useAuth/useRoles.ts
```diff
@@ -36,11 +37,27 @@ function getRoleListFromJwt(jwtString?: AccessToken) {
...
+}
+
 export const useRoles = (accessToken?: AccessToken) => {
   const roles = useMemo(() => {
     return getRoleListFromJwt(accessToken);
   }, [accessToken]);
 
+  const githubUserId = useMemo(() => {
+    if (accessToken && roles.includes(HasuraUserRole.RegisteredUser)) {
+      try {
```
ofux, line 52: it looks good to me to have a try/catch block, but why `ledProjectIds` doesn't have it for example?

ofux (approved):

frontend/src/hooks/useAuth/useRoles.ts
AnthonyBuisset: It has, but it is hidden => https://github.com/onlydustxyz/marketplace/pull/561/files/53ba0dd3400d8912349ae6f329ab02b101fd3983#diff-0dbff201018139acabae1f54aa4fe9019d935577659bae7227a608f4d3a4a40fR21-R38

This function will decode JWT once and, only if succeed, check for the project lead role. So if we have the role, then the decoding succeeded.

However, the registered user role is set only if JWT is not empty, even if not valid.

PR #563, opened 2023-01-23
Support multiple PATs in Github API client

ofux:
- ✨ support multiple PATs in Github API client

### DIFF
--- a/backend/common/infrastructure/src/github/mod.rs
+++ b/backend/common/infrastructure/src/github/mod.rs
@@ -1,11 +1,15 @@
-use std::{collections::HashMap, fmt::Debug};
+use std::{
+	collections::HashMap,
+	fmt::Debug,
+	sync::{Arc, Mutex},
+};
 
 use anyhow::anyhow;
 use octocrab::{
 	models::{pulls::PullRequest, repos::Content, Repository, User},
 	FromResponse, Octocrab, OctocrabBuilder,
 };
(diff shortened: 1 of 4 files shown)


backend/common/infrastructure/src/github/mod.rs
```diff
@@ -21,20 +25,57 @@ mod specifications;
...
 	pub fn new(config: &Config) -> anyhow::Result<Self> {
-		let instance = Octocrab::builder()
-			.base_url(&config.base_url)?
-			.personal_token(config.personal_access_token.clone())
-			.add_headers(&config.headers)?
-			.build()?;
-		Ok(Self(instance))
+		let personal_access_tokens: Vec<&str> = config
+			.personal_access_tokens
+			.split(',')
```
AnthonyBuisset, line 41: why not using a Vec directly in the yaml ?

AnthonyBuisset (approved):

backend/common/infrastructure/src/github/mod.rs
AnthonyBuisset: nevermind, it is easier to handle like this from github secrets

ofux: Exactly. I didn't want to be obliged to define multiple env vars (one per token). I also didn't want to limit the token count because of the yaml config.

PR #569, opened 2023-01-24
💄 display "Payout info missing" in My Payments table

ofux:
- 💄 display "Payout info missing" in My Payments table

Closes E-252

Note: the "Payment info missing" status is intentionally not designed properly as it will be done in another PR.

### DIFF
--- a/frontend/src/components/PayoutTable/PayoutTable.stories.tsx
+++ b/frontend/src/components/PayoutTable/PayoutTable.stories.tsx
@@ -36,6 +36,12 @@ const mockPayments: Payment[] = [
   },
 ];
 
-const Template: ComponentStory<typeof Payments> = () => <Payments payments={mockPayments} />;
+const Template: ComponentStory<typeof Payments> = args => (
+  <Payments payments={mockPayments} payoutInfoMissing={args.payoutInfoMissing} />
+);
 
 export const Default = Template.bind({});
+
+Default.args = {
+  payoutInfoMissing: false,
+};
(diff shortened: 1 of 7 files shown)


frontend/src/components/PayoutTable/PayoutTable.stories.tsx
```diff
@@ -35,6 +35,12 @@ const mockPayments: Payment[] = [
   },
 ];
 
-const Template: ComponentStory<typeof Payments> = () => <Payments payments={mockPayments} />;
+const Template: ComponentStory<typeof Payments> = args => (
+  <Payments payments={mockPayments} payoutInfoMissing={args.payoutInfoMissing} />
```
AnthonyBuisset, line 39: Can you add a control for this new arg so that we can play with it in the storybook ? Like here: https://github.com/onlydustxyz/marketplace/blob/main/frontend/src/components/ProjectCard/ProjectCard.stories.tsx

AnthonyBuisset: Storybook => https://63984b27c30bb6c931960c17-vcdqmlkoln.chromatic.com/?path=/story/paymenttable--default

AnthonyBuisset (approved):

frontend/src/components/PayoutTable/PayoutTable.stories.tsx
ofux: this is here: https://63984b27c30bb6c931960c17-vcdqmlkoln.chromatic.com/?path=/story/payments--default
Not sure to understand the difference between those 2 components

ofux: oh wait, "Payment" is actually the "PayoutTable", which is different than the "PaymentTable". The component's name in storybook is misleading

PR #571, opened 2023-01-24
Update GitHub connect button styles

Bernardstanislas:
Closes E-208


frontend/src/icons/GithubLogo.tsx
```diff
@@ -0,0 +1,7 @@
+interface Props {
+  className?: string;
```
ofux, line 2: @oscarwroche made a comment on one of my PRs on this specific point. I just copy-paste it here:

**Standard: [*Tailwind class name*](https://www.notion.so/onlydust/Tailwind-class-name-e9762becd8324d959ca9127a7205914a#8a86b0a416aa4984bc4607a0d14d126d)**
Nitpicking here, but it's better to have an explicit class name (like 'variant') rather than a generic className prop

ofux (approved):

PR #573, opened 2023-01-24
:boom: nullable profile fields

oscarwroche:


frontend/src/pages/Profile/components/ProfileForm.tsx
```diff
@@ -312,17 +313,13 @@ const mapFormDataToSchema = ({
...
-    optBankAddress: payoutType === PayoutSettingsType.BankAddress && IBAN && BIC ? { IBAN, BIC } : null,
-    optEthName: payoutType === PayoutSettingsType.EthereumName ? ethIdentity : null,
   };
 
-  const variables: UpdateProfileInfoMutationVariables = { email, identity, location, payoutSettings };
-
-  return {
-    variables,
-  };
+  return { variables: { email, identity, location, payoutSettings } };
```
AnthonyBuisset, line 322: it seems you are always sending objects (even if empty), so maybe the mutation can request mandatory values ?

PR #576, opened 2023-01-25
Make Only Dust logo clickable

ofux:
- 💄 Make Only Dust logo clickable


oscarwroche (approved):

oscarwroche:
@ofux fix / get rid of Only Dust logo test and it's good to go

PR #578, opened 2023-01-25
Use patched Octocrab version that supports cache

Bernardstanislas:
This is a first attempt at bringing cache to Octocrab.

This uses a forked version of Octocrab, which includes [http-cache-reqwest](https://github.com/06chaynes/http-cache/tree/latest/http-cache-reqwest) with an in-memory cache manager – Heroku's filesystem is readonly, so we cannot go with [CaCache](https://github.com/06chaynes/http-cache/tree/latest/http-cache-reqwest#features).

## What's next

- check that `http-cache` works well with Github Etags
- follow [the discussion aiming at removing `reqwest` from Octocrab](https://github.com/XAMPPRocky/octocrab/issues/99)
- fix everything I broke in Octocrab by making proper PRs to its dependencies:
  - in the patched version, Octocrab does not have a retry strategy anymore 🙇

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -24,7 +24,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "5c192eb8f11fc081b0fe4259ba5af04217d4e0faddd02417310a927911abd7c8"
 dependencies = [
  "crypto-common",
- "generic-array",
+ "generic-array 0.14.6",
 ]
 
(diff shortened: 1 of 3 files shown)


oscarwroche (approved):

Bernardstanislas:
Not relevant, let's put a cache in the reverse proxy directly and suggested by @ofux

PR #581, opened 2023-01-26
:lipstick: contributors table header styling

oscarwroche:


frontend/src/components/Table/HeaderCell.tsx
```diff
@@ -1,18 +1,24 @@
...
+  width?: HeaderCellWidth;
 }
 
-export const HeaderCell: React.FC<Props> = ({ className, onClick, children }) => {
+export const HeaderCell: React.FC<Props> = ({ onClick, children, leftMargin, width }) => {
   return (
     <th
       scope="col"
-      className={`px-3 text-left h-full ${onClick && "hover:cursor-pointer"} ${className}`}
+      className={`${leftMargin && "px-3"} text-left h-full ${onClick && "hover:cursor-pointer"} ${width}`}
```
AnthonyBuisset, line 18: px means horizontal margin, not necessarily left
Also I suggest to turn this className into `classNames` as there are already 2 booleans in it

AnthonyBuisset (approved):

PR #583, opened 2023-01-26
💄 Do not display "My Payments" in navbar when the user has no payment

ofux:
Closes E-229


.eslintrc.js
```diff
@@ -44,5 +44,12 @@ module.exports = {
     ],
     "@typescript-eslint/no-namespace": "off",
   },
-  ignorePatterns: ["frontend/dist/*", "frontend/test/*", "cypress/*", "app.json", "scripts/*"],
+  ignorePatterns: [
+    "frontend/dist/*",
+    "frontend/test/*",
+    "frontend/src/__generated",
```
AnthonyBuisset, line 50: ❤️

AnthonyBuisset (changes requested):
Can you please remove the PaymentTableFallback as well as it does not fit the design and should never appear.

I would go even further and make the route protected so that you cannot go directly to the page using the url `https://app.onlydust.xyz/my-contributions`

PR #584, opened 2023-01-26
Filter projects by technology

AnthonyBuisset:
- 🔧 move local cache control into default values
- 💄 do not display technologies tag when empty
- :sparkles: create Badge component
- :sparkles: create FilterDropDown component
- ✨ create FilterPanel component
- ♻️ move ProjectList into dedicated component
- ♻️ factorize project card fragment
- ✨ add filter panel in browse projects page
- ✅ add e2e test


cypress/e2e/projects_browsing.cy.js
```diff
@@ -0,0 +1,21 @@
+describe("As a visitor, I", () => {
+    const OFUX = 595505;
+    const ANTHO = 990474;
```
ofux, line 3: I like seeing my github handle in the repo, but it looks like those constants are not used :)

ofux (approved):

cypress/e2e/projects_browsing.cy.js
AnthonyBuisset: indeed, wrong copy paste :)

PR #586, opened 2023-01-27
:construction: new cta in table

oscarwroche:


Bernardstanislas (approved):

frontend/src/components/Button/index.tsx
```diff
@@ -1,7 +1,9 @@
+import classNames from "classnames";
 import { PropsWithChildren } from "react";
 
 export enum ButtonSize {
   Small = "small",
+  Medium = "medium",
   Large = "large",
 }
```
tdelabro, line 8: Why do you use strings as enum? Rather than default behavior, counting from 0.
You always use them like this `if size === ButtonSize.Large`, never accessing the inner value.
Allocating whole strings when you can just an int sound like a huge waste to me.

tdelabro (approved):

frontend/src/components/Button/index.tsx
oscarwroche: https://www.typescriptlang.org/docs/handbook/enums.html#string-enums

C'est mieux pour le debugging entres autres

PR #591, opened 2023-01-27
🗃️ drop legacy tables

AnthonyBuisset:


ofux (approved):

ofux (changes requested):
There is a relationship between users and payout settings that needs to be removed

ofux (approved):

PR #596, opened 2023-01-29
Add the hours range to the estimation

Bernardstanislas:
Closes E-287
Closes E-288


frontend/src/hooks/useWorkEstimation.ts
```diff
@@ -102,6 +102,13 @@ export const useWorkEstimation = (
     onChange(amountToPay);
   }, [amountToPay]);
 
+  const decreaseButtonDisabled = useMemo(() => {
```
AnthonyBuisset: I would rather call it "canDecrease" as this hook is not suppose to know there is a button.

AnthonyBuisset (approved):

PR #597, opened 2023-01-30
Fix my payments table sorting

ofux:
- 🐛 Fix my payments table sorting

### DIFF
--- a/frontend/src/components/PayoutTable/index.tsx
+++ b/frontend/src/components/PayoutTable/index.tsx
@@ -56,12 +56,11 @@ const PayoutTable: React.FC<PropsType> = ({ payments, payoutInfoMissing }) => {
         case Field.Date:
           return payment.requestedAt;
         case Field.Contribution: {
-          const repoName = payment.reason.match(REPO_NAME);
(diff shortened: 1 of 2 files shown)


frontend/src/components/PayoutTable/index.tsx
```diff
@@ -56,12 +56,11 @@ const PayoutTable: React.FC<PropsType> = ({ payments, payoutInfoMissing }) => {
         case Field.Date:
           return payment.requestedAt;
         case Field.Contribution: {
-          const repoName = payment.reason.match(REPO_NAME);
           const issueNumber = payment.reason.match(ISSUE_NUMBER);
-          return `${repoName}${issueNumber}`;
+          return `${payment.project.title}${issueNumber}`;
```
AnthonyBuisset, line 60: maybe we should use localeCompare for string comparison

ofux: please look my last commit. The sorting is a bit more verbose, but indeed better.

AnthonyBuisset (approved):

PR #603, opened 2023-01-30
sort languages to get the most used only

ofux:
- 🐛 sort languages to get the most used only

### DIFF
--- a/frontend/src/utils/languages.test.ts
+++ b/frontend/src/utils/languages.test.ts
@@ -1,0 +1,48 @@
+import { describe, expect, it } from "vitest";
+import { getMostUsedLanguages } from "./languages";
+
+describe("Languages", () => {
+  it("should return most used languages in descending order", () => {
+    const result = getMostUsedLanguages(
+      {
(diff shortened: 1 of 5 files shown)


AnthonyBuisset (approved):

frontend/src/utils/languages.test.ts
```diff
@@ -0,0 +1,48 @@
+import { describe, expect, it } from "vitest";
+import { getMostUsedLanguages } from "./languages";
+
+describe("Languages", () => {
+  it("should return most used languages in descending order", () => {
```
Bernardstanislas, line 5: You can even use https://vitest.dev/api/#test-each for this :)

Bernardstanislas (approved):

frontend/src/utils/languages.test.ts
ofux: oh that's right, it would have been better here. Making the change...

ofux: done

PR #606, opened 2023-01-30
feat: do not display send payment button in tables when user not proj…

tdelabro:
…ect leader

fix: typo


Bernardstanislas (commented):
I think you should regenerated the schema :)
It also seems you tried to rename `ledProjectIds` to `leadProjectIds`, but the `led` is intentional here, it is the passive form of "lead"

tdelabro:
>  It also seems you tried to rename ledProjectIds to leadProjectIds, but the led is intentional here, it is the passive form of "lead"

ooooooooooooooooh

Bernardstanislas (approved):

PR #612, opened 2023-01-31
User redirection at sign-in and sign-up

AnthonyBuisset:
closes E-231


frontend/src/__generated/graphql.tsx
```diff
@@ -2584,7 +2584,9 @@ export type GetPaymentRequestIdsQuery = { __typename?: 'query_root', paymentRequ
 
 export type ProjectCardFieldsFragment = { __typename?: 'Projects', id: any, name: string, totalSpentAmountInUsd: any, projectDetails: { __typename?: 'ProjectDetails', description: string | null, telegramLink: string | null, logoUrl: string | null } | null, pendingInvitations: Array<{ __typename?: 'PendingProjectLeaderInvitations', id: any }>, projectLeads: Array<{ __typename?: 'ProjectLeads', user: { __typename?: 'users', displayName: string, avatarUrl: string } | null }>, githubRepo: { __typename?: 'GithubRepoDetails', name: string, owner: string, languages: any, content: { __typename?: 'Repository', logoUrl: string, contributors: Array<{ __typename?: 'User', login: string, avatarUrl: string }> } } | null };
 
-export type PendingProjectLeaderInvitationsQueryVariables = Exact<{ [key: string]: never; }>;
+export type PendingProjectLeaderInvitationsQueryVariables = Exact<{
+  githubUserId: InputMaybe<Scalars['bigint']>;
```
Bernardstanislas, line 2555: 😆

frontend/src/hooks/useSession.tsx
```diff
@@ -1,39 +1,55 @@
...
+  SetLastVisitedProjectId = "lastVisitedProjectId",
+}
+
+type Action = {
+  method: SessionMethod;
+  value: string;
+};
+
+function reduce(state: Session, action: Action): Session {
+  return { ...state, [action.method]: action.value };
```
Bernardstanislas, line 58: this looks weird to me, why would you use such a reducer?

Bernardstanislas: ok I think I get what you are doing here, the names are confusing to me.
The `action` is not really an action, it is a `key` in a key/value store, right?
I would change the type `Action` with the following:
```
type Entry = {
  key: SessionKey;
  value: string;
};
```

AnthonyBuisset: it is actually an action, to modify a given key in the session object. The first version of the code was looking like that:
```
switch(action.method) {
   case SessionMethod.SetLastVisitedProjectId: {
      return { ...session, lastVisitedProjectId: action.value };
   }
   case SessionMethod.SetLastLoginTime: {
      return { ...session, lastLoginTime: action.value };
   }
   case SessionMethod.SetVisitedPageBeforeLogin: {
      return { ...session, vistedPageBeforeLogin: action.value };
   }
}
```

which is nicely refactored as:
```
return { ...session, [action.method]: action.value };
```

if the method string value is modified accordingly

Bernardstanislas: Ok as seen IRl, I get it

Bernardstanislas (approved):

PR #613, opened 2023-01-31
Style the pull request input with the correct error message

ofux:
- 💄 improve design of form Input and Github PR link card
- 💄 design error banner for github PR input field
- ♻️ 💄 refactor translation keys and design contributor card

### DIFF
--- a/frontend/src/components/FormInput/View.tsx
+++ b/frontend/src/components/FormInput/View.tsx
@@ -48,7 +48,7 @@ const View: React.FC<PropsType> = ({
           placeholder={placeholder}
           type={type}
           className={classNames(
-            "w-full h-11 bg-white/5 border border-greyscale-50/[0.08] rounded-xl font-walsheim font-normal text-base px-4 py-3 placeholder:text-greyscale-50/60 focus:placeholder:text-spacePurple-200/60 focus:outline-double focus:outline-spacePurple-500 focus:border-spacePurple-500 focus:bg-spacePurple-900",
+            "w-full h-11 bg-white/5 border border-greyscale-50/[0.08] rounded-xl font-walsheim font-normal text-base px-4 py-3 placeholder:text-greyscale-400 focus:placeholder:text-spacePurple-200/60 focus:outline-double focus:outline-spacePurple-500 focus:border-spacePurple-500 focus:bg-spacePurple-900",
             { "border outline-1 outline-rose-600 border-rose-600": error && errorType === InputErrorType.Normal }
           )}
           value={value}
--- a/frontend/src/components/FormInput/View.tsx
+++ b/frontend/src/components/FormInput/View.tsx
@@ -47,8 +47,10 @@ const View: React.FC<PropsType> = ({
           id={name}
           placeholder={placeholder}
           type={type}
-          className={`w-full h-11 bg-white/5 border border-greyscale-50/[0.08] rounded-xl font-walsheim font-normal text-base px-4 py-3 placeholder:text-greyscale-50/60 focus:placeholder:text-spacePurple-200/60 focus:outline-double focus:outline-spacePurple-500 focus:border-spacePurple-500 focus:bg-spacePurple-900
-          ${error && errorType === InputErrorType.Normal && "border outline-1 outline-rose-600 border-rose-600"}`}
+          className={classNames(
+            "w-full h-11 bg-white/5 border border-greyscale-50/[0.08] rounded-xl font-walsheim font-normal text-base px-4 py-3 placeholder:text-greyscale-50/60 focus:placeholder:text-spacePurple-200/60 focus:outline-double focus:outline-spacePurple-500 focus:border-spacePurple-500 focus:bg-spacePurple-900",
+            { "border outline-1 outline-rose-600 border-rose-600": error && errorType === InputErrorType.Normal }
+          )}
           value={value}
           {...register}
           onChange={onChange}
@@ -58,7 +60,7 @@ const View: React.FC<PropsType> = ({
       {errorType === InputErrorType.Normal && (
         <span className="text-rose-600 text-sm ml-3">{error?.message ? error.message.toString() : "\u00A0"}</span>
       )}
-      {error && errorType === InputErrorType.Banner && error.message && (
+      {error?.message && errorType === InputErrorType.Banner && (
         <div className="flex">
           <ImageCard
             backgroundImageUrl={headerElementBackground}
--- a/frontend/src/components/FormInput/View.tsx
+++ b/frontend/src/components/FormInput/View.tsx
@@ -38,7 +38,7 @@ const View: React.FC<PropsType> = ({
   register,
   onChange,
 }) => (
-  <label html-for={name} className="flex flex-col flex-grow gap-2 text-greyscale-50/70 font-walsheim">
+  <label html-for={name} className="flex flex-col flex-grow gap-2 text-greyscale-300 font-walsheim">
     <div className="font-medium text-sm tracking-tight">{label}</div>
     <div className={classNames("flex flex-col", { "gap-8": errorType === InputErrorType.Banner })}>
       <div className="relative flex items-center">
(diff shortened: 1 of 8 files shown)


frontend/src/components/FormInput/View.tsx
```diff
@@ -37,14 +44,19 @@ const View: React.FC<PropsType> = ({
           placeholder={placeholder}
           type={type}
           className={`w-full h-11 bg-white/5 border border-greyscale-50/[0.08] rounded-xl font-walsheim font-normal text-base px-4 py-3 placeholder:text-greyscale-50/60 focus:placeholder:text-spacePurple-200/60 focus:outline-double focus:outline-spacePurple-500 focus:border-spacePurple-500 focus:bg-spacePurple-900
-          ${error && "border outline-1 outline-rose-600 border-rose-600"}`}
+          ${error && errorType === InputErrorType.Normal && "border outline-1 outline-rose-600 border-rose-600"}`}
```
AnthonyBuisset, line 51: good opportunity to use `classNames`

frontend/src/components/ImageCard/index.tsx
```diff
@@ -71,6 +86,9 @@ export default function ImageCard({
             "bg-noise-light": backgroundNoise === BackgroundNoise.Light,
             "bg-noise-medium": backgroundNoise === BackgroundNoise.Medium,
             "bg-noise-heavy": backgroundNoise === BackgroundNoise.Heavy,
+          },
+          {
+            "backdrop-blur-4xl": backgroundBlur === BackgroundBlur.Heavy,
```
AnthonyBuisset, line 91: you are duplicating the css value here with the enum value

AnthonyBuisset (approved):

frontend/src/components/FormInput/View.tsx
```diff
@@ -24,11 +30,12 @@ const View: React.FC<PropsType> = ({
   value,
   loading,
   error,
+  errorType,
   register,
   onChange,
 }) => (
-  <label html-for={name} className="flex flex-col flex-grow gap-2 text-greyscale-50 font-walsheim">
-    <div className="font-medium text-sm">{label}</div>
+  <label html-for={name} className="flex flex-col flex-grow gap-2 text-greyscale-50/70 font-walsheim">
```
Bernardstanislas, line 41: `text-greyscale-50/70` looks weird to me, we don't want transparency on texts. Maybe you should just use another color in the palette and notify Gaetan that the color is artificially created in figma

frontend/src/components/FormInput/View.tsx
```diff
@@ -37,14 +44,19 @@ const View: React.FC<PropsType> = ({
           placeholder={placeholder}
           type={type}
           className={`w-full h-11 bg-white/5 border border-greyscale-50/[0.08] rounded-xl font-walsheim font-normal text-base px-4 py-3 placeholder:text-greyscale-50/60 focus:placeholder:text-spacePurple-200/60 focus:outline-double focus:outline-spacePurple-500 focus:border-spacePurple-500 focus:bg-spacePurple-900
```
Bernardstanislas, line 50: Refactor to `classNames` for clarity here :)

frontend/src/components/FormInput/View.tsx
```diff
@@ -37,14 +44,19 @@ const View: React.FC<PropsType> = ({
...
           {...register}
           onChange={onChange}
         />
         {loading && <LoaderIcon className="flex animate-spin place-items-center absolute right-0 mr-3" />}
       </div>
-      <span className="text-rose-600 text-sm ml-3">{error?.message ? error.message.toString() : "\u00A0"}</span>
+      {errorType === InputErrorType.Normal && (
+        <span className="text-rose-600 text-sm ml-3">{error?.message ? error.message.toString() : "\u00A0"}</span>
+      )}
+      {error && errorType === InputErrorType.Wide && error.message && (
```
Bernardstanislas: `error && error?.message ?` is equivalent to `error?.message ?`

frontend/src/components/FormInput/View.tsx
```diff
@@ -37,14 +44,19 @@ const View: React.FC<PropsType> = ({
...
+        <span className="text-rose-600 text-sm ml-3">{error?.message ? error.message.toString() : "\u00A0"}</span>
```
Bernardstanislas: You don't need to check for `error?.message?` existence here since you did it the line before

frontend/src/pages/Profile/components/ProfileRadioGroup/View.tsx
```diff
@@ -14,7 +14,7 @@ type PropsType = {
 export default function View({ label, options, register }: PropsType) {
   return (
     <label className="flex flex-col flex-grow gap-2 mb-6">
-      <div className="font-medium text-sm text-neutral-100">{label}</div>
+      <div className="font-medium text-sm text-greyscale-50/70 tracking-tight">{label}</div>
```
Bernardstanislas, line 17: Same here, text color should be 100% opacity

frontend/src/components/FormInput/View.tsx
```diff
@@ -54,8 +58,25 @@ const View: React.FC<PropsType> = ({
       {errorType === InputErrorType.Normal && (
         <span className="text-rose-600 text-sm ml-3">{error?.message ? error.message.toString() : "\u00A0"}</span>
       )}
-      {error && errorType === InputErrorType.Wide && error.message && (
-        <span className="text-rose-600 text-sm ml-3">{error?.message ? error.message.toString() : "\u00A0"}</span>
+      {error && errorType === InputErrorType.Banner && error.message && (
+        <div className="flex">
+          <ImageCard
+            backgroundImageUrl={headerElementBackground}
+            backgroundPosition={BackgroundPosition.TopLeft}
```
Bernardstanislas, line 65: Why topLeft and not just top or center?

Bernardstanislas (changes requested):

frontend/src/components/ImageCard/index.tsx
ofux: I don't understand

frontend/src/components/FormInput/View.tsx
ofux: It's honestly a nightmare to implement the design. Either we respect Figma and then it is not acceptable from a technical point of view (like here), or we respect the technical point of view and it doesn't respect the design anymore.

frontend/src/components/FormInput/View.tsx
ofux: done

frontend/src/components/FormInput/View.tsx
ofux: was fixed in the next commit :)

frontend/src/components/FormInput/View.tsx
ofux: Because otherwise I don't get the same background as shown in Figma

frontend/src/components/FormInput/View.tsx
ofux: I choose `text-greyscale-300` instead of `text-greyscale-50/70` as I think it is the closest color we have.

Bernardstanislas: Yes, for the font colors we have an agreement with Gaetan, we need to notify him every time we face some inconsistencies

Bernardstanislas (approved):

frontend/src/components/ImageCard/index.tsx
AnthonyBuisset: "backdrop-blur-4xl" is the value of `BackgroundBlur.Heavy`
so I suggest either to put a more descriptive value in the enum (e.g. `Heavy = "heavy")
or to use directly the value of the enum in the classname (e.g. `backgroundBlur: backgroundBlur !== BackgroundBlur.None`)

ofux: oh right! good catch!

frontend/src/components/FormInput/View.tsx
ofux: ok, I'll know it for the next time :)

PR #619, opened 2023-01-31
Prefill the payment form input with github handle

AnthonyBuisset:
closes E-282

### DIFF
--- a/frontend/src/components/ContributorsTable/index.tsx
+++ b/frontend/src/components/ContributorsTable/index.tsx
@@ -5,9 +5,11 @@ import {
   ProjectDetailsDispatchContext,
 } from "src/pages/ProjectDetails/ProjectDetailsContext";
 import { rates } from "src/hooks/useWorkEstimation";
+import { generatePath, useNavigate } from "react-router-dom";
 import { gql } from "@apollo/client";
 import { ContributorsTableFieldsFragment } from "src/__generated/graphql";
-import View from "./View";
+import View, { Contributor } from "./View";
+import { RoutePaths } from "src/App";
 
 type PropsType = {
   contributors: ContributorsTableFieldsFragment[];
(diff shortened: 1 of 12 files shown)


frontend/src/pages/ProjectDetails/PaymentActions/PaymentForm/hooks/useIsGithubLoginValid/index.tsx
```diff
@@ -0,0 +1,31 @@
+import { gql } from "@apollo/client";
```
Bernardstanislas, line 1: Move it to the shared `hooks` folder

frontend/src/hooks/useHasuraQuery.tsx
```diff
@@ -18,6 +20,14 @@ export const useHasuraQuery = <T, V = OperationVariables>(
   return useQuery(query, merge(options, { context: { headers: { "X-Hasura-Role": role } } }));
 };
 
+export const useHasuraLazyQuery = <T, V = OperationVariables>(
```
Bernardstanislas, line 23: We faced a lot of troubles with @oscarwroche with the lazy queries, especially when loging out/in, is everything running smooth for you?

frontend/src/components/ContributorsTable/index.tsx
```diff
@@ -158,11 +163,14 @@ const renderContributors = (contributors: Contributor[], isProjectLeader: boolea
...
           <div
             onClick={() => {
-              !isSendingNewPaymentDisabled &&
+              if (!isSendingNewPaymentDisabled) {
                 dispatch({
                   type: ProjectDetailsActionType.SelectPaymentAction,
                   selectedPaymentAction: PaymentAction.Send,
                 });
+                query.set("githubUserLogin", contributor.login);
+                navigate({ ...location, search: `?${query.toString()}` });
```
Bernardstanislas, line 172: I think we should leverage more React router for this, see https://reactrouter.com/en/main/route/route#optional-segments

Bernardstanislas (changes requested):

frontend/src/hooks/useHasuraQuery.tsx
AnthonyBuisset: so far so good

frontend/src/components/ContributorsTable/index.tsx
AnthonyBuisset: will check !

frontend/src/pages/ProjectDetails/PaymentActions/PaymentForm/hooks/useIsGithubLoginValid/index.tsx
AnthonyBuisset: done

frontend/src/components/ContributorsTable/index.tsx
AnthonyBuisset: Looking at the documentation, using the location state is even better IMO. 
It is persisted during a window.reload but not part of the URL.
I think it fits really nicely here as the login is neither a search param, neither a nested route (like `project/:projectId`)

Bernardstanislas (approved):

PR #623, opened 2023-02-01
E 259 add date to payments table

oscarwroche:


frontend/src/components/PaymentTable/index.tsx
```diff
@@ -55,6 +69,7 @@ const renderHeaders = () => {
 const renderPayments = (payments: PaymentRequest[]) => {
   return payments.map(payment => (
     <Line key={payment.id} highlightOnHover={200}>
+      <Cell> {dayjs.tz(payment.requestedAt, dayjs.tz.guess()).fromNow()} </Cell>
```
Bernardstanislas, line 72: Can we refactor this to put in common with "my payments" table? This is exactly the same logic

Bernardstanislas (changes requested):

Bernardstanislas (approved):

PR #643, opened 2023-02-02
Refresh projections and sanity checks binaries

AnthonyBuisset:
closes E-300
closes E-329

### DIFF
--- a/backend/api/src/bin/events_sanity_checks/main.rs
+++ b/backend/api/src/bin/events_sanity_checks/main.rs
@@ -1,0 +1,36 @@
+use std::{collections::HashSet, sync::Arc};
+
+use anyhow::Result;
+use api::Config;
+use domain::{AggregateEvent, EventSourcable, EventStore, Payment, Project};
+use dotenv::dotenv;
+use infrastructure::{
+	config,
+	database::{init_pool, Client as DatabaseClient},
+	tracing::Tracer,
+};
+
+#[tokio::main]
+async fn main() -> Result<()> {
+	dotenv().ok();
+	let config: Config = config::load("backend/api/app.yaml")?;
+	let _tracer = Tracer::init(config.tracer(), "events_sanity_checks")?;
+
+	let database = Arc::new(DatabaseClient::new(init_pool(config.database())?));
+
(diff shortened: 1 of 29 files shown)


backend/api/src/bin/events_sanity_checks/main.rs
```diff
@@ -0,0 +1,36 @@
...
+use domain::{AggregateEvent, EventSourcable, EventStore, Payment, Project};
+use dotenv::dotenv;
+use infrastructure::{
+	config,
+	database::{init_pool, Client as DatabaseClient},
+	tracing::Tracer,
+};
+
+#[tokio::main]
+async fn main() -> Result<()> {
```
Bernardstanislas, line 14: Where will we be using it? Do you consider having it in the CI for example?

backend/event-listeners/README.md
```diff
@@ -3,6 +3,21 @@
...
--   The presentation layer (REST API)
--   The application layer (use cases triggered by the application layer)
--   End-to-end tests as a standalone binary
+- The presentation layer (REST API)
+- The application layer (use cases triggered by the application layer)
+- End-to-end tests as a standalone binary
+
+## Refresh projections
+
+To refresh any aggregate projections, simply run the following command:
```
Bernardstanislas, line 12: I think you should also document how to do it on remote envs like staging or production. I guess we'll be using https://devcenter.heroku.com/articles/one-off-dynos

Bernardstanislas (approved):

backend/api/src/bin/events_sanity_checks/main.rs
AnthonyBuisset: It is part of https://linear.app/onlydust/issue/E-330/automate-sanity-checks-in-staging-and-production
I think it will be run after each staging load if possible and triggered by the `promote-to-production` script

backend/event-listeners/README.md
AnthonyBuisset: indeed, it is planned but I need to load it before to play with the CLI.
I have added a todo in https://linear.app/onlydust/issue/E-330/automate-sanity-checks-in-staging-and-production

PR #644, opened 2023-02-02
Display the user avatar in the contributor selector

Bernardstanislas:


frontend/src/components/FormInput/View.tsx
```diff
@@ -20,6 +20,7 @@ type PropsType = {
   onFocus?: FocusEventHandler<unknown>;
   prefixComponent?: React.ReactNode;
   suffixComponent?: React.ReactNode;
+  inputClassName?: string;
```
AnthonyBuisset, line 23: this is not following standard https://www.notion.so/onlydust/Frontend-component-3d254726c7604fe7866abd2966f9eaef#75c7437d6b714040993e09bacee4e00e

Bernardstanislas: I know but I had no choice, the input cannot know the width of the prefix element, so the caller has to control it with a padding class

PR #661, opened 2023-02-07
Replace payment requests subscription with query

AnthonyBuisset:
- ♻️ move usePayoutSettings to dedicated hook
- ♻️ use query instead of subscription
- ✨ update query cache after mutation success


frontend/src/pages/ProjectDetails/PaymentActions/index.tsx
```diff
@@ -39,7 +40,7 @@ export default function PaymentActions({ projectId }: PaymentsProps) {
               {payments.length > 0 ? (
                 <Card>
                   <Suspense fallback={<Loader />}>
-                    <PaymentTable payments={payments} />
+                    <PaymentTable payments={cloneDeep(payments)} />
```
Bernardstanislas, line 43: Why do you need to clone here?

Bernardstanislas (approved):

frontend/src/pages/ProjectDetails/PaymentActions/index.tsx
AnthonyBuisset: reliqua from my many trials, well spotted !

PR #670, opened 2023-02-09
e 367 projections are not properly refreshed

AnthonyBuisset:
- ♻️ move payment event inside budget
- ♻️ move budget logic inside project aggregate
- ♻️ payment is no longer an aggregate root
- 🔥 remove project total spent amount projection
- ♻️ update event listeners
- ♻️ update refresh binary
- ⚰️ remove Budget::Spent event
- ♻️ update usecases and mutations
- 🚨 fix clippy warnings
- 🗃️ migrate events
- ✅ fix e2e tests
- 🛂 add public permissions on budgets
- ♻️ use budgets to compute total spent amount
- 🗃️ remove total spent amount from projects projection
- 🏷️ update codegen


backend/common/domain/src/budget/aggregate.rs
```diff
@@ -49,15 +72,22 @@ impl Aggregate for Budget {
 }
 
 impl EventSourcable for Budget {
-	fn apply_event(self, event: &Self::Event) -> Self {
+	fn apply_event(mut self, event: &Self::Event) -> Self {
```
Bernardstanislas, line 104: I think it would be worth justifying this choice, we can imagine having a pure `apply_event` function that would make a copy of the payments too?

backend/common/domain/src/event.rs
```diff
@@ -28,26 +27,16 @@ mod test {
 	use serde_json::{json, Value};
 
 	use super::*;
-	use crate::{Amount, BlockchainNetwork, Currency, PaymentReceipt};
 
 	#[test]
 	fn display_event_as_json() {
-		let event = Event::Payment(PaymentEvent::Processed {
+		let event = Event::Project(ProjectEvent::LeaderAssigned {
```
Bernardstanislas, line 33: You don't want to test the `PaymentEvent::Processed` event anymore? Might be coming in a future commit 👀

frontend/src/components/ProjectCard/index.tsx
```diff
@@ -96,7 +99,14 @@ export const PROJECT_CARD_FRAGMENT = gql`
   fragment ProjectCardFields on Projects {
     id
     name
-    totalSpentAmountInUsd
+    budgetsAggregate {
```
Bernardstanislas, line 102: amazing!!

backend/api/Procfile
```diff
@@ -1,2 +1,3 @@
 web: ROCKET_PORT=$PORT ./target/release/api
 release: ./hasura/update-metadata.sh
+events_sanity_checks: ./target/release/events_sanity_checks
```
Bernardstanislas, line 3: Make sure you untoggle them in the Heroku interface as it will run them by default

Bernardstanislas (approved):

backend/common/domain/src/budget/aggregate.rs
AnthonyBuisset: `mut self` is taking ownership of the memory and returning it by value (different than `&mut self`). So the result is exactly the same as copying the payments. The only impact here is performance.

backend/api/Procfile
AnthonyBuisset: sure !

backend/common/domain/src/event.rs
AnthonyBuisset: we could indeed test with `ProjectEvent::BudgetEvent::PaymentEvent::Processed` but the test is not verifying values anyway, just that it does not fail while serializing in json

backend/api/Procfile
AnthonyBuisset: we would like this dyno to be run automatically but to be shutdown after each run, we will play with it in staging

PR #674, opened 2023-02-10
E 369

tdelabro:


AnthonyBuisset (changes requested):
LGTM

AnthonyBuisset (approved):

backend/api/src/presentation/http/dto/ethereum.rs
```diff
@@ -0,0 +1,42 @@
...
+pub enum EthereumIdentityType {
+	EthereumAddress,
+	EthereumName,
+}
+
+impl TryFrom<EthereumIdentityInput> for EthereumIdentity {
+	type Error = anyhow::Error;
+
+	fn try_from(value: EthereumIdentityInput) -> Result<Self, Self::Error> {
+		match value.r#type {
```
Bernardstanislas, line 23: I think this is not flexible enough, we want to have a single string field that is checked to be either a valid ethereum address, or an ens.
The tools calling the mutation don't know how to distinguish one from the other, so we cannot expect them to provide a `type`

Bernardstanislas (changes requested):

backend/api/src/presentation/http/dto/ethereum.rs
tdelabro: I cannot really do that without moving in the mutation itself some logic that belongs to the GraphQl layer.
And without removing the strong types we have in this API.
I would rather have the front adapt. It can detect that, either by providing the users with two forms, or by doing a first parse of the user input

Bernardstanislas (approved):

PR #681, opened 2023-02-13
✨ add feedback form

ofux:


frontend/src/components/Button/index.tsx
```diff
@@ -13,6 +13,11 @@ export enum ButtonType {
   Secondary = "secondary",
 }
 
+export enum Border {
+  Normal = "normal",
+  Less = "less",
```
Bernardstanislas, line 18: Maybe use "Small" to have the same scale as with other variants in other components?

Bernardstanislas (approved):

cypress.config.js
```diff
@@ -28,4 +28,5 @@ module.exports = defineConfig({
   },
   video: false,
   defaultCommandTimeout: 15000,
+  chromeWebSecurity: false,
```
ofux, line 31: Note: this is required to test iframe content within cypress

Bernardstanislas (approved):

PR #682, opened 2023-02-13
B 358 fix unwanted component reloads

oscarwroche:


Bernardstanislas (approved):
I see where this is going :)

PR #692, opened 2023-02-14
feat(db): add sponsors and projects_sponsors table

tdelabro:


migrations/2023-02-14-140832_add-sponsors/up.sql
```diff
@@ -0,0 +1,12 @@
+CREATE TABLE sponsors (
+    id UUID PRIMARY KEY DEFAULT gen_random_uuid(),
+    name TEXT UNIQUE,
+    logo_url TEXT
+);
+
+CREATE TABLE projects_sponsors (
+    id UUID PRIMARY KEY DEFAULT gen_random_uuid(),
+    project_id UUID REFERENCES projects(id),
```
Bernardstanislas, line 9: can you put a cascade delete? Because what will happen, is that when we refresh the projection of a project aggregate, the `project` db model will be deleted, and without cascading it will fail because of this foreign key :)

Bernardstanislas (changes requested):

migrations/2023-02-14-140832_add-sponsors/up.sql
tdelabro: done

Bernardstanislas (approved):

PR #700, opened 2023-02-16
B 685 remove timer based token reload

oscarwroche:


frontend/src/components/ApolloWrapper/index.tsx
```diff
@@ -27,6 +31,112 @@ const ApolloWrapper: React.FC<PropsWithChildren> = ({ children }) => {
...
+      });
+      return tokenSetResponse.data;
+    },
+    handleFetch: tokenSetResponse => {
+      if (setTokenSet) {
+        setTokenSet({ ...tokenSetResponse, creationDate: new Date() });
+      }
+    },
+    handleResponse: (operation, accessTokenField) => (tokenSetResponse: TokenSet) => {
+      if (tokenSetResponse.accessToken) {
```
ofux, line 53: What happens if it's false? Maybe we should log an error in the `else`?

frontend/src/components/ApolloWrapper/index.tsx
```diff
@@ -27,6 +31,112 @@ const ApolloWrapper: React.FC<PropsWithChildren> = ({ children }) => {
...
+          headers: {
+            ...headers,
+            authorization: tokenSetResponse.accessToken ? `Bearer ${tokenSetResponse.accessToken}` : "",
+          },
+        }));
+      }
+      return { [accessTokenField]: tokenSetResponse };
+    },
+    handleError: error => {
+      console.error(error);
```
ofux, line 64: 👍 food for datadog

ofux (approved):

frontend/src/components/ApolloWrapper/index.tsx
oscarwroche: Yes and we should probably log out too

oscarwroche: Log out will be automatically happen with `handleError` if an error is thrown in `handleResponse`

oscarwroche: @ofux fixed in latest commit

Bernardstanislas (approved):

PR #702, opened 2023-02-16
feat: remplace project description by short and long descriptions

tdelabro:


backend/api/src/application/project/create.rs
```diff
@@ -39,9 +39,10 @@ impl Usecase {
 		name: NonEmptyTrimmedString,
 		initial_budget: Amount,
 		github_repo_id: GithubRepositoryId,
-		description: Option<String>,
 		telegram_link: Option<String>,
 		logo_url: Option<String>,
+		short_description: String,
+		long_description: String,
```
ofux, line 45: `NonEmptyTrimmedString` would be better here to validate inputs

ofux (changes requested):
Just add a bit of input validation (see comment) and fix e2e tests :)

backend/api/src/application/project/create.rs
tdelabro: done

ofux (approved):

PR #707, opened 2023-02-17
Introduce new mutation `unlink_github_repo`

ofux:
- ✨ Introduce new mutation `unlink_github_repo`
- ✅ add integration test
- ✅ improve integration test to check short and long descriptions

### DIFF
--- a/backend/api/src/infrastructure/database/project_github_repo.rs
+++ b/backend/api/src/infrastructure/database/project_github_repo.rs
@@ -17,8 +17,11 @@ mockall::mock! {
 	pub Repository {
 		pub fn new(client: Arc<Client>) -> Self;
 		pub fn try_insert(&self, id1: &<Project as domain::Entity>::Id, id2: &<GithubRepo as domain::Entity>::Id)  -> Result<(), infrastructure::database::DatabaseError>;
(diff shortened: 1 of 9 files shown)


backend/api/src/infrastructure/database/project_github_repo.rs
```diff
@@ -17,8 +17,11 @@ mockall::mock! {
 	pub Repository {
 		pub fn new(client: Arc<Client>) -> Self;
 		pub fn try_insert(&self, id1: &<Project as domain::Entity>::Id, id2: &<GithubRepo as domain::Entity>::Id)  -> Result<(), infrastructure::database::DatabaseError>;
+		pub fn delete(&self, id1: &<Project as domain::Entity>::Id, id2: &<GithubRepo as domain::Entity>::Id)  -> Result<(), infrastructure::database::DatabaseError>;
 		#[allow(non_snake_case)]
 		pub fn delete_all_GithubRepos_of(&self, id1: &<Project as domain::Entity>::Id)  -> Result<(), infrastructure::database::DatabaseError>;
+		#[allow(non_snake_case)]
```
Bernardstanislas: Can we do something about it in the macro generating the actual repo?

Bernardstanislas (approved):

backend/api/src/infrastructure/database/project_github_repo.rs
ofux: sure, we just need to transform CamelCase to snake_case, there should be some crates to do so.
I'll do it in another PR ;)

PR #717, opened 2023-02-21
New mutations to add/remove project sponsors

AnthonyBuisset:
closes B-730


ofux (approved):
LGTM. I was just wondering if we want to check the existence of the project and the sponsor in `add_sponsor` usecase... ? WDYT?

AnthonyBuisset:
> 

I do not think it is necessary as we already have foreign keys.

PR #722, opened 2023-02-22
🗑️ clean-up SHOW_SPONSORS feature flag

AnthonyBuisset:


oscarwroche:
@AnthonyBuisset We should add a story for project cards with sponsors

oscarwroche (changes requested):
We should add a story for cards with sponsors and update / add tests

PR #731, opened 2023-02-24
Remove initial budget from create_project mutation

AnthonyBuisset:
closes B-714

### DIFF
--- a/backend/common/domain/src/project/aggregate.rs
+++ b/backend/common/domain/src/project/aggregate.rs
@@ -1,4 +1,4 @@
-use std::{collections::HashSet, iter::once};
+use std::collections::HashSet;
 
 use derive_getters::{Dissolve, Getters};
 use derive_more::Constructor;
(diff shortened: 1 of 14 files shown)


backend/common/domain/src/project/aggregate.rs
```diff
@@ -74,14 +75,20 @@ impl Project {
...
-			.allocate(*diff)?
+	pub fn allocate_budget(&self, diff: &Amount) -> Result<Vec<<Self as Aggregate>::Event>> {
+		let events = match self.budget.as_ref() {
+			Some(budget) => budget.allocate(*diff.amount())?,
+			None => {
+				let events = Budget::create(BudgetId::new(), diff.currency().clone());
+				let budget = Budget::from_events(&events);
+				events.into_iter().chain(budget.allocate(*diff.amount())?).collect()
+			},
+		};
```
ofux, line 86: Something looks wrong to me. This method doesn't receive a `diff` anymore, but the `remaining_amount`.
Either this method should be changed, or the caller should compute the diff like it used to do (it will need to handle the case where there is no budget yet)

It looks like e2e tests confirm that: `expected 300 to equal 0`

ofux (changes requested):

backend/common/domain/src/project/aggregate.rs
AnthonyBuisset: Indeed, I fixed it but maybe a conflict removed my fix. I will re-do it.

ofux (approved):

PR #737, opened 2023-02-27
In Browse projects, aggregate technologies and contributors across repositories

ofux:
- ✨ aggregate contributors and languages from multiple repos
- 🚧 temp build fix, waiting for the project details page to be updated
- ✅ update e2e tests

### DIFF
--- a/frontend/src/components/ProjectCard/index.tsx
+++ b/frontend/src/components/ProjectCard/index.tsx
@@ -10,10 +10,15 @@ import { viewportConfig } from "src/config";
 import { useIntl } from "src/hooks/useIntl";
 import CodeSSlashLine from "src/icons/CodeSSlashLine";
 import { Project } from "src/pages/Projects";
-import { buildLanguageString } from "src/utils/languages";
+import { buildLanguageString, getDeduplicatedAggregatedLanguages } from "src/utils/languages";
 import { formatMoneyAmount } from "src/utils/money";
 import { useMediaQuery } from "usehooks-ts";
-import { ProjectLeadFragmentDoc, SponsorFragmentDoc } from "src/__generated/graphql";
+import {
+  ProjectCardContributorsFieldsFragment,
+  ProjectCardGithubRepoFieldsFragment,
+  ProjectLeadFragmentDoc,
+  SponsorFragmentDoc,
+} from "src/__generated/graphql";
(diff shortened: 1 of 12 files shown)


frontend/src/utils/languages.ts
```diff
@@ -21,3 +23,34 @@ export const buildLanguageString = (languageMap: LanguageMap) => {
...
+
+export const getDeduplicatedAggregatedLanguages = function (
+  githubRepos: GithubRepoLanguagesFieldsFragment[] | undefined
+): LanguageMap {
+  if (githubRepos === undefined) {
+    return {};
+  }
+  return githubRepos
+    .map(repo => repo.githubRepoDetails?.languages)
+    .filter((languages): languages is LanguageMap => !!languages) // ⚠️ runtime type guard
```
AnthonyBuisset, line 35: you can use `filter(isDefined)`

frontend/src/components/ProjectCard/index.tsx
```diff
@@ -135,9 +145,48 @@ export default function ProjectCard({
...
 
+export const getDeduplicatedAggregatedContributors = function (
+  githubRepos: ProjectCardGithubRepoFieldsFragment[] | undefined
+): ProjectCardContributorsFieldsFragment[] {
+  if (githubRepos === undefined) {
+    return [];
+  }
+  const flatten_users = githubRepos
+    .flatMap(repo => repo.githubRepoDetails?.content?.contributors)
+    .flatMap(user => (user ? [user] : []));
```
AnthonyBuisset, line 156: you can use `filter(isDefined)`

AnthonyBuisset (approved):

frontend/src/utils/languages.ts
ofux: done

frontend/src/components/ProjectCard/index.tsx
ofux: in this specific case, I'm pretty sure this is not strictly equivalent. The version with flatMap is (I think) safer in terms of typing because the `isDefined` returns `argument is T` which - if I understood well - can only be checked at runtime by typescript.

AnthonyBuisset: that's above my knowledge ^^

AnthonyBuisset (approved):

PR #738, opened 2023-02-27
In Browse projects, update metadata

ofux:
- ⚰️ remove useless fields in PROJECT_CARD_CONTRIBUTORS_FRAGMENT
- ✨ show repo count on project cards
- ✅ update e2e test

### DIFF
--- a/frontend/src/components/ProjectCard/Tag/Tag.stories.tsx
+++ b/frontend/src/components/ProjectCard/Tag/Tag.stories.tsx
@@ -1,0 +1,19 @@
+import { ComponentStory, ComponentMeta } from "@storybook/react";
+import GitRepositoryLine from "src/icons/GitRepositoryLine";
+
+import Tag from ".";
+
+export default {
+  title: "Tag",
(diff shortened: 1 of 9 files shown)


frontend/src/components/ProjectCard/Tag/Tag.stories.tsx
```diff
@@ -0,0 +1,19 @@
+import { ComponentStory, ComponentMeta } from "@storybook/react";
```
AnthonyBuisset, line 1: we already have a Tag component, could it be re-used ?
`src/components/Tag`

AnthonyBuisset (approved):

frontend/src/components/ProjectCard/Tag/Tag.stories.tsx
ofux: nope, it's quite different, it's the tags in the payment table. I'd rather move the `src/components/Tag` component to be a child of the payment table.

PR #739, opened 2023-02-27
[B-690] Show all repositories in project details page

AnthonyBuisset:
closes B-690
closes B-689


oscarwroche:
There seems to be a display problem when the description is not long enough 
Screenshot 2023-02-27 at 18 57 36

oscarwroche:
There seems to be a case when the contributor name display does not work ("github.com/undefined") @AnthonyBuisset 
Screenshot 2023-02-27 at 19 00 25

oscarwroche (changes requested):
See visual edges cases in this thread / in Chromatic

AnthonyBuisset:
> There seems to be a display problem when the description is not long enough @AnthonyBuisset Screenshot 2023-02-27 at 18 57 36

This happens only when no repo, we should never be the case as the project won't be visible. But fixed anyway 👍

AnthonyBuisset:
> There seems to be a case when the contributor name display does not work ("github.com/undefined") @AnthonyBuisset Screenshot 2023-02-27 at 19 00 25

This should never happen as well but fixed anyway 👍

oscarwroche (approved):

PR #743, opened 2023-02-28
Use an external service to store images

ofux:
- ✨ add image_store service
- ♻️ use image_store service in create project mutation

### DIFF
--- a/backend/api/src/infrastructure/simple_storage/mod.rs
+++ b/backend/api/src/infrastructure/simple_storage/mod.rs
@@ -4,7 +4,7 @@ use std::{
 	str::FromStr,
 };
 
-use anyhow::anyhow;
+use anyhow::{anyhow, Result};
 use async_trait::async_trait;
 use rusoto_core::Region;
 use rusoto_s3::{PutObjectRequest, S3Client, StreamingBody, S3};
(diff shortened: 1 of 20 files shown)


backend/api/src/infrastructure/simple_storage/mod.rs
```diff
@@ -0,0 +1,112 @@
...
+pub struct Client {
+	s3_client: S3Client,
+	images_bucket_name: String,
+}
+
+impl Client {
+	pub async fn new(config: &Config) -> Self {
+		let s3_client = S3Client::new(
+			Region::from_str(config.bucket_region.as_str())
+				.unwrap_or_else(|_| panic!("Invalid AWS region {}", config.bucket_region)),
```
AnthonyBuisset, line 33: I would rather rename the function to `build` and return a.`Result` to let the caller panic if needed. Here we do not control where we are called from.

backend/api/src/infrastructure/simple_storage/mod.rs
```diff
@@ -0,0 +1,112 @@
...
+		);
+
+		// Check credentials as soon as the client is created
+		s3_client
+			.list_buckets()
+			.await
+			.expect("Could not list buckets. Check your AWS credentials.");
```
AnthonyBuisset, line 40: 1. why listing buckets if not used ?
2. returns a `Result` as above

AnthonyBuisset (approved):

backend/api/src/infrastructure/simple_storage/mod.rs
ofux: Agreed

backend/api/src/infrastructure/simple_storage/mod.rs
ofux: It's a way to fail-fast in case the AWS credentials are incorrect. The sdk doesn't check it until a real call is made, and I think it is better to fail as soon as the Client is created.

AnthonyBuisset (approved):

PR #758, opened 2023-03-03
b-754-as-a-user-i-want-my-filtering-options-to

oscarwroche:


frontend/src/pages/Projects/index.tsx
```diff
@@ -14,27 +14,37 @@ export enum ProjectOwnershipType {
...
+  technologies: string[];
+}
+
 export default function Projects() {
   const { T } = useT();
   const { ledProjectIds } = useAuth();
 
-  const [technologies, setTechnologies] = useState<string[]>([]);
-  const [projectOwnershipType, setProjectOwnershipType] = useState(ProjectOwnershipType.All);
+  const [projectFilter, setProjectFilter] = useLocalStorage<ProjectFilter>("PROJECT_FILTER", {
```
AnthonyBuisset, line 26: I have created a useSession hook to abstract the useLocalStorage, maybe we can use it here ? (and rename it as it is not using session storage...)

AnthonyBuisset (approved):

frontend/src/pages/Projects/FilterPanel/index.tsx
```diff
@@ -40,9 +42,18 @@ export const GET_ALL_TECHNOLOGIES_QUERY = gql`
   query GetAllTechnologies {
     projects {
```
AnthonyBuisset, line 43: you could use `...VisibleProjectFragment` syntax to stay synchronized with future changes

AnthonyBuisset (approved):

PR #783, opened 2023-03-08
[B-794] Update payout settings validity computation

AnthonyBuisset:
closes B-794


ofux (approved):
Just clean the schema.rs and it's all good

PR #786, opened 2023-03-09
[B-158] Handle errors coming from remote schemas in Hasura

AnthonyBuisset:
closes B-158

### DIFF
--- a/backend/github-proxy/src/presentation/graphql/query.rs
+++ b/backend/github-proxy/src/presentation/graphql/query.rs
@@ -1,7 +1,7 @@
 use domain::GithubRepositoryId;
 use juniper::{graphql_object, DefaultScalarValue};
 
-use super::{Context, Result};
+use super::Context;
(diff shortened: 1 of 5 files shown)


backend/github-proxy/src/presentation/graphql/query.rs
```diff
@@ -16,38 +19,45 @@ impl Query {
...
+}
+
+trait Logged {
+	fn logged(self) -> Self;
+}
+
+impl<T, E: Error> Logged for Result<T, E> {
+	fn logged(self) -> Self {
+		if let Err(error) = &self {
+			error!("{error}");
```
ofux, line 59: I think it would be better to log NotFound errors at WARN level (as it was done before) otherwise we might have false-positive production alerts

AnthonyBuisset: indeed, done

ofux (approved):

PR #807, opened 2023-03-13
Improve datadog rum intake

Bernardstanislas:
Now using our brand new reverse proxy location for sending sessions to datadog!


ofux (approved):
LGTM, however I wonder what could happen to our `https://gateway.onlydust.xyz` URL if it starts getting blocked by some ad-blockers because of this.
We'll see but maybe we'll have to use a distinct domain just to avoid being black listed.

Bernardstanislas:
> LGTM, however I wonder what could happen to our `https://gateway.onlydust.xyz` URL if it starts getting blocked by some ad-blockers because of this. We'll see but maybe we'll have to use a distinct domain just to avoid being black listed.

I don't think ad blockers block a whole domain based on the content sent to it but I'm not sure neither

PR #824, opened 2023-03-20
[E-460] Restore scroll position when navigating back to projects

AnthonyBuisset:
closes E-460
closes B-809


Bernardstanislas (approved):
LGTM, I wonder if you tried to use React router's scroll restoration feature? https://reactrouter.com/en/main/components/scroll-restoration
Looks like your solution works, but maybe using react router is more idiomatic

PR #833, opened 2023-03-22
:bug: fix overflow-hidden after typeform submission

oscarwroche:


frontend/src/pages/Payments/InvoiceSubmission/View.tsx
```diff
@@ -21,35 +22,67 @@ type Props = {
 export default function InvoiceSubmission({ paymentRequests, githubUserId, markInvoiceAsReceived, userInfos }: Props) {
   const { T } = useIntl();
 
+  const [isSubmitted, setIsSubmitted] = useState(false);
+  const [isClosed, setIsClosed] = useState(false);
```
AnthonyBuisset, line 26: should not we reset isSubmitted and isClosed at some point ?
like onOpen and after the mutation is a success ?

AnthonyBuisset (approved):

PR #857, opened 2023-04-04
Add hours_worked parameter in Payment::Requested events

ofux:
- ✨ send hoursWorked to request payment mutation
- ✨ add duration_worked in Payment::Requested events and projection


AnthonyBuisset (approved):
I do not think it was necessary to project the duration worked, but it is available now for future use

PR #864, opened 2023-04-05
Impersonate users using admin access

Bernardstanislas:
Closes b-840


ofux (commented):
LGTM, but don't you think a little e2e test would be nice to test impersonation? Like a simple scenario in which you impersonate a contributor and see its payments.

Bernardstanislas:
> LGTM, but don't you think a little e2e test would be nice to test impersonation? Like a simple scenario in which you impersonate a contributor and see its payments.

Sure, on it!

ofux (approved):

PR #880, opened 2023-04-14
✨ [E-471] Publish payment request info on GitHub

oscarwroche:


backend/api/src/infrastructure/github/service.rs
```diff
@@ -84,8 +84,10 @@ impl GithubService for github::Client {
 			.state(models::IssueState::Closed)
 			.send()
 			.await
-			.map_err(Into::<github::Error>::into)?;
-
+			.map_err(|err| {
```
AnthonyBuisset, line 87: why do we need a warning here ?

AnthonyBuisset (approved):

PR #908, opened 2023-04-24
[E-512] Contributors page and select take too long to display

AnthonyBuisset:
closes E-512


ofux (approved):

backend/event-listeners/src/domain/projections/github_pulls.rs
```diff
@@ -0,0 +1,24 @@
...
+}
+
+impl domain::Entity for GithubPull {
+	type Id = GithubIssueId;
+}
+
+#[derive(Default, Debug, Clone, AsChangeset, Serialize, Deserialize, Constructor)]
+#[table_name = "github_pulls"]
+pub struct GithubMergedPull {
+	merged_at: Option<NaiveDateTime>,
```
Bernardstanislas: Why is it an `Option`?

Bernardstanislas (approved):

backend/event-listeners/src/domain/projections/github_pulls.rs
AnthonyBuisset: Indeed, also this struct as been removed in a later commit

PR #944, opened 2023-05-02
Install the app once a day

Bernardstanislas:


.github/workflows/install.yml
```diff
@@ -0,0 +1,24 @@
+name: Install
+
+on:
+  schedule:
+    - cron: "0 */2 * * *"
```
AnthonyBuisset, line 5: that's not once a day 😁

Bernardstanislas: Hahaha indeed I copied the lighthouse file

AnthonyBuisset (approved):

PR #967, opened 2023-05-10
💄 fix responsiveness in case of long project names

AnthonyBuisset:


Bernardstanislas (approved):
That's great, but can we take the opportunity to have a test for this? I can think about it in another PR if you don't have an obvious solution right now

PR #989, opened 2023-05-16
💄 default sorting order depends on column

ofux:


AnthonyBuisset (approved):

AnthonyBuisset:
In E-523 it is specified to save user's preference in local storage. Is it planned in another PR ?

PR #991, opened 2023-05-17
🐛 make sure issues' author are indexed and marked as contributor of the issue's repo

ofux:


ofux:
In the end, we think this is a bad idea: this would make anyone who simply open an issue or a PR a contributor. We decide to stick to the github's contributor definition (ie. someone who made at least one commit on the repo).

PR #1003, opened 2023-05-22
[E-543] Allow contributors to show their interest to projects

AnthonyBuisset:
closes E-543


oscarwroche (approved):

backend/common/domain/src/project/aggregate.rs
```diff
@@ -238,6 +239,21 @@ impl Project {
 			.map(|event| ProjectEvent::Budget { id: self.id, event })
 			.collect())
 	}
+
+	pub fn apply(
```
ofux, line 243: I think we should check that there is not already an application from the same user here.

AnthonyBuisset: I was wondering if we should allow a user to apply several times. I will a check

ofux (approved):

PR #1011, opened 2023-05-24
[E-513] Fix tooltip display under card

AnthonyBuisset:
closes E-513

Following [react-tooltip recommendations](https://react-tooltip.com/docs/troubleshooting#dynamically-generated-anchor-elements) I moved all tooltips to use the same single instance.

I am then creating a portal to render the tooltip to make sure it is always on top of everything.

### DIFF
--- a/frontend/src/components/Tooltip/index.tsx
+++ b/frontend/src/components/Tooltip/index.tsx
@@ -8,24 +8,32 @@ export enum TooltipPosition {
   Right = "right",
 }
 
-type Props = {
-  anchorId: string;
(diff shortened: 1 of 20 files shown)


frontend/src/components/Tooltip/index.tsx
```diff
@@ -29,7 +30,7 @@ export default function Tooltip({
   anchorSelect,
   children,
 }: TooltipProps) {
-  return (
+  return createPortal(
```
oscarwroche, line 33: 🤯

oscarwroche (approved):

PR #1023, opened 2023-05-25
✨  [E-538] Accept terms and conditions

oscarwroche:


backend/api/src/presentation/graphql/mutation.rs
```diff
@@ -338,6 +338,14 @@ impl Mutation {
 		Ok(caller_id.into())
 	}
 
+	pub async fn accept_tc(context: &Context) -> Result<Uuid> {
```
AnthonyBuisset, line 341: as discussed, can you please rename at least the mutation to be more explicit as this is our pubilc API ?

PR #1055, opened 2023-06-15
🐛 fix user selected to be indexed

ofux:


backend/event-listeners/src/bin/github-indexer/indexer/user.rs
```diff
@@ -43,6 +44,14 @@ impl State {
 		Self {
 			hash: hash(user),
 			last_updated_time: Utc::now(),
+			last_indexed_time: Utc::now(),
+		}
+	}
+
+	fn touch(self) -> Self {
```
AnthonyBuisset, line 51: I would name it `touched` as you return self by value

AnthonyBuisset (approved):

PR #1069, opened 2023-06-22
🗃️ rationalize indexes

ofux:


migrations/2023-06-21-235927_add-indexes/up.sql
```diff
@@ -0,0 +1,36 @@
+CREATE UNIQUE INDEX IF NOT EXISTS "github_repos_contributors_user_id_repo_id_idx" ON public.github_repos_contributors (user_id, repo_id);
+
+
+CREATE INDEX IF NOT EXISTS "github_issues_repo_id_issue_number_idx" ON public.github_issues (repo_id, issue_number);
+
+
+CREATE INDEX IF NOT EXISTS "github_issues_type_status_idx" ON public.github_issues ("type", "status");
+
+
+CREATE INDEX IF NOT EXISTS "github_issues_created_at_idx" ON public.github_issues (created_at);
```
AnthonyBuisset, line 10: shouldn't this be a DESC index ?

AnthonyBuisset (approved):
Did you check we already have indexes on our different project sorting ? project name, spent budget, contributor count, etc ?

migrations/2023-06-21-235927_add-indexes/up.sql
ofux: I don't think so, I didn't see any `order by created_at` anywhere. Some `min(created_at)` and `max(created_at)` but that's all

PR #1187, opened 2023-08-22
Resize project logo url with cloudflare

PierreOucif:


alexbensimon (approved):

ofux (approved):
If you are sur 100px is enough, then it LGTM

PR #1204, opened 2023-08-30
move event listeners into a dedicated binary

AnthonyBuisset:


backend/event-listeners/src/main.rs
```diff
@@ -10,7 +9,5 @@ async fn main() -> Result<()> {
 	let config: Config = config::load("backend/event-listeners/app.yaml")?;
 	let _tracer = Tracer::init(config.tracer.clone(), "event-queue-worker")?;
 
-	try_join_all(bootstrap(config).await?).await?;
```
ofux, line 13: Are we obliged to keep this file?

AnthonyBuisset: it will be used very soon

PR #1237, opened 2023-09-17
[E-743] Refactor common code for budget allocation

AnthonyBuisset:


backend/common/domain/src/aggregate/pending.rs
```diff
@@ -0,0 +1,40 @@
...
+			pending_events: Default::default(),
+		}
+	}
+}
+
+impl<A: super::Aggregate> Iterator for Aggregate<A> {
+	type Item = A::Event;
+
+	fn next(&mut self) -> Option<Self::Item> {
+		self.pending_events.pop_front()
```
ofux: Isn't it error prone? Iterating on pending events also means removing them from the vec. Is it something considered idiomatic and safe?

AnthonyBuisset: this is standard behavior of `vec.into_iter()` as it takes self by value
if we want to iterate without removing items, we need `vec.iter()` that returns references.
But as we do not need it yet, it is not implemented

ofux (approved):
Well done!

PR #1262, opened 2023-09-25
[E 759] Set pending lead invite project card background

kaelsky:
Capture d’écran 2023-09-25 à 11 54 46

- [x] Add orange-800 color to tailwind theme
- [x] Set pending lead project card invite bg colors
- [x] Add 20px top margin when pending invite project is the first element of the list

### DIFF
--- a/frontend/src/components/ProjectCard/index.tsx
+++ b/frontend/src/components/ProjectCard/index.tsx
@@ -25,10 +25,10 @@ export type Project = ArrayElement<GetProjectsQuery["projects"]>;
 
 type ProjectCardProps = {
   project: Project;
-  isFirst?: boolean;
+  className?: string;
 };
 
-export default function ProjectCard({ project, isFirst }: ProjectCardProps) {
+export default function ProjectCard({ project, className }: ProjectCardProps) {
   const {
     id,
     key,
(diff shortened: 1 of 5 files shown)


frontend/src/pages/Projects/AllProjects/index.tsx
```diff
@@ -94,7 +94,7 @@ export default function AllProjects({
       </div>
       <div className="flex grow flex-col gap-5">
         {projects && projects.length > 0 ? (
-          projects.map(project => <ProjectCard key={project.id} project={project} />)
+          projects.map((project, index) => <ProjectCard key={project.id} project={project} isFirst={index === 0} />)
```
haydencleary, line 97: You could replace the ternary condition with `projects?.map()`, couldn't you ?

frontend/src/components/ProjectCard/index.tsx
```diff
@@ -54,15 +55,17 @@ export default function ProjectCard({ project }: ProjectCardProps) {
 
   const card = (
     <Card
-      className={classNames("relative bg-noise-light hover:bg-right", {
-        "xl:bg-orange-500/8 xl:hover:bg-orange-500/12": pendingInvitations?.length > 0,
+      className={classNames("relative hover:bg-right", {
+        "bg-noise-light": pendingInvitations?.length === 0,
+        "xl:bg-orange-900 xl:hover:bg-noise-light ": pendingInvitations?.length > 0,
+        "mt-3": isFirst && pendingInvitations?.length > 0,
```
haydencleary, line 61: I don't think spacing styles should be handled inside the component. I recommend this sort of styling be handled by the parent or wrap the component if necessary.

kaelsky: Good point, I agree, context styling styling should be handled in context

haydencleary (approved):

PR #1273, opened 2023-10-02
[E-798] Disabling rewards when no budget

kaelsky:
Capture d’écran 2023-10-02 à 17 12 38

- [x] Disabled rewards buttons when not enough budget
- [x] Display tooltips when buttons are disabled

### DIFF
--- a/src/pages/ProjectDetails/Contributors/ContributorsTable/View.tsx
+++ b/src/pages/ProjectDetails/Contributors/ContributorsTable/View.tsx
@@ -46,7 +46,7 @@ export default function View({
   remainingBudget,
   onRewardGranted: onPaymentRequested,
 }: Props) {
-  const isSendingNewPaymentDisabled = remainingBudget < rates.hours;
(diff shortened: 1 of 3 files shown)


src/pages/ProjectDetails/Contributors/ContributorsTable/View.tsx
```diff
@@ -46,7 +46,7 @@ export default function View({
   remainingBudget,
   onRewardGranted: onPaymentRequested,
 }: Props) {
-  const isSendingNewPaymentDisabled = remainingBudget < rates.hours;
+  const isSendingNewPaymentDisabled = remainingBudget < rates.hours || remainingBudget === 0;
```
AnthonyBuisset, line 49: I think we need to define a minimal threshold. We use to consider not enough budget if less than 1h. But for multi currencies we need to see with Gregoire for a rule

AnthonyBuisset (approved):

PR #1278, opened 2023-10-03
Refactor/GitHub icons

haydencleary:
I removed the duplicate SVG icons I added in the contribution list PR.
I also improved how some of the other icons are styled, now they use less classes and are more predicatable.


kaelsky (approved):
LGTM ! Attention a bien check les chromatics pour éviter les regressions

PR #1281, opened 2023-10-04
[E 627] New contributions date tooltips

kaelsky:
- [x] Update commits tooltip font-size 
- [x] Tooltips on dates & status
- [x] Open User profil side panel


src/components/Contribution/ContributionTable.tsx
```diff
@@ -231,6 +232,7 @@ export function ContributionTable({
               type={contribution.type as GithubContributionType}
               status={draft ? GithubPullRequestDraft.Draft : (contributionStatus as GithubItemStatus)}
               date={new Date(lineDate)}
+              tooltipVariant={Variant.Default}
```
haydencleary, line 235: Should be the `Blue` variant here.

src/components/GithubCard/utils.ts
haydencleary, line 1: Perhaps this would be better off in the utils directory ? Something like `utils/date.ts`.
Then maybe rename the functions to something more explicit. For the date function, as it is in a British format you could rename it to `getFormattedDateGB` or something. Same for the American time format.

src/components/GithubCard/GithubPullRequest/GithubPullRequest.tsx
```diff
@@ -30,6 +25,18 @@ export enum Action {
...
 }
 
+function getPullRequestStatusDate(pullRequest: GithubPullRequestWithCommitsFragment) {
+  switch (pullRequest.status) {
+    case GithubPullRequestStatus.Closed:
+      return new Date(pullRequest.closedAt);
+    case GithubPullRequestStatus.Merged:
+      return new Date(pullRequest.mergedAt);
+    default:
+    case GithubPullRequestStatus.Open:
```
haydencleary, line 35: I'd put the default case last.

src/components/GithubCard/GithubPullRequest/CommitsTooltip.tsx
```diff
@@ -18,31 +16,23 @@ export function CommitsTooltip({
   contributorLogin,
 }: CommitsTooltipProps) {
   const { T } = useIntl();
+
   return (
-    <div className="flex flex-col">
+    <div className="flex flex-col gap-1">
       <>
```
haydencleary, line 22: Unnecessary fragment.

src/components/Contribution/ContributionDate.tsx
```diff
@@ -11,53 +10,76 @@ import {
...
-  };
+  const formattedTime = new Intl.DateTimeFormat("en-US", { timeStyle: "short" }).format(new Date(date));
 
   return (
     <>
       <Tooltip id={tooltipId} clickable position={TooltipPosition.Top} variant={tooltipVariant}>
         <div className="flex items-center gap-2 px-1 py-2">
           <ContributionIcon type={type} status={status} />
-          <p className="text-sm">
+          <p>
```
haydencleary, line 82: For the contribution table, the date tooltip font size is too small. Maybe due to this class change ?

src/components/Contribution/ContributionDate.tsx
```diff
@@ -11,53 +10,76 @@ import {
...
-      [GithubIssueStatus.Open]: "contributions.tooltip.dateAssigned",
-      [GithubIssueStatus.Completed]: "contributions.tooltip.dateClosed",
-      [GithubIssueStatus.Cancelled]: "contributions.tooltip.dateClosed",
-    },
-    [GithubContributionType.CodeReview]: {
-      [GithubCodeReviewStatus.Pending]: "contributions.tooltip.dateOpened",
-      [GithubCodeReviewStatus.Completed]: "contributions.tooltip.dateClosed",
-    },
-  };
+  const formattedTime = new Intl.DateTimeFormat("en-US", { timeStyle: "short" }).format(new Date(date));
```
haydencleary, line 75: Maybe use the new utils here ?

src/components/Contribution/ContributionDate.tsx
```diff
@@ -66,9 +88,18 @@ export function ContributionDate({
...
+        {withIcon ? (
+          <>
+            <ContributionIcon type={type} status={status} size={withIcon ? Sizes.xs : undefined} />
+            {T(tokensShort[type][status as keyof typeof tokensShort[GithubContributionType]], {
+              date: displayRelativeDate(date),
+            })}
+          </>
+        ) : (
+          displayRelativeDate(date)
+        )}
```
haydencleary, line 101: I'm not sure about the `withIcon` prop. The case where `withIcon` is true, a text token is also rendered.  I wonder if it would be better to pass children instead ? That way we can render whatever we want depending on the case. What do you think ?

haydencleary (changes requested):
I can see in the 3 different Github card types there is duplicate UI, perhaps it would be wise to extract it ?

haydencleary (approved):

PR #1286, opened 2023-10-05
[E-632] add timestamp to contributions empty states

kaelsky:
- [x] Display last repos indexetion timestamp in contributions sidepanel EmptyStates


haydencleary (approved):
lgtm, there just seems to be an error here `.storybook/stories/EmptyState.stories.tsx`

PR #1291, opened 2023-10-09
[E-810] Style update of reward item tooltips

kaelsky:
- [x] Tooltips have now bottom position
- [x] Remove padding
- [x] Text XS
- [x] Global 100ms delay for all tooltips

### DIFF
--- a/src/components/Contribution/ContributionDate.tsx
+++ b/src/components/Contribution/ContributionDate.tsx
@@ -13,6 +13,7 @@ import displayRelativeDate from "src/utils/displayRelativeDate";
 import { ContributionIcon, Sizes } from "./ContributionIcon";
 import { getFormattedDateGB, getFormattedTimeUS } from "src/utils/date";
 import { ComponentProps } from "react";
+import { cn } from "src/utils/cn";
 
 const tokens: GithubTypeStatusDict<string> = {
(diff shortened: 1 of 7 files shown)


src/components/Contribution/ContributionDate.tsx
```diff
@@ -52,29 +53,35 @@ const tokensShort: GithubTypeStatusDict<string> = {
...
+  withIcon = false,
+  tooltipProps = {
+    position: TooltipPosition.Top,
+    variant: Variant.Blue,
+  },
+}: ContributionDateProps) {
   const { T } = useIntl();
 
   const tooltipId = `${id}-date-tooltip`;
+  const { className, ...rest } = tooltipProps;
```
haydencleary, line 79: I find it strange to extract the `className` prop from `tooltipProps` to then apply it to the children. Shouldn't it be a separate prop ?

kaelsky: It's `className` from `tooltipProps` object equivalent to `tooltipProps.className`

kaelsky: We will refactor this tooltip behavior (https://linear.app/onlydust/issue/B-1025/[refacto]-tooltip-component)

haydencleary (approved):

PR #1333, opened 2023-10-30
Add new tailwind components

alexbeno:


pixelfact (approved):

.storybook/stories/Typography.stories.mdx
haydencleary, line 1: Storybook doesn't seem to be rendering the typography elements correctly.

Screenshot 2023-10-30 at 09 48 34

tailwind.config.ts
```diff
@@ -220,6 +232,111 @@ const config: Config = {
           };
           return acc;
         }, {}),
+        /* -------------------------------------------------------------------------- */
+        /*                                     NEW                                    */
+        /* -------------------------------------------------------------------------- */
+
+        /* -------------------------------- TYPESCALE ------------------------------- */
+        ".text-title-xl": {
```
haydencleary, line 240: It might be clearer to define these with the `@apply` syntax. That way you could group repeated attributes like `font-family`, `border` etc.

.storybook/stories/Typography.stories.mdx
alexbeno: Yes, I think it's because the classes are not yet used anywhere in the code, so Tailwind doesn't compile them. When we use one, Storybook updates.

tailwind.config.ts
alexbeno: The only issue, but it might be a configuration error in my editor,

If I move these classes into a CSS file with the "component" layer, do we lose autocomplete? I don't know everything about Tailwind yet; do you know if it's still possible to have autocomplete?

.storybook/stories/Typography.stories.mdx
haydencleary: It may be possible to tell Tailwind to compile classes found inside storybook too

tailwind.config.ts
haydencleary: I think it should work, you could try a quick test. If you get autocomplete it may be worth refactoring.
It's not a deal breaker, it just might avoid duplication.

alexbeno: totally agree i will try

.storybook/stories/Typography.stories.mdx
alexbeno: I will try !

PR #1345, opened 2023-10-31
Currency component - E-696 && E-695

alexbeno:
- Reduce font-weight for total in$ it's too bold
- dollar equivalent tootip should be displayed (and centered) when hovering dollars, not other currency amount 
- can you increase font size for dollar equivalent within tooltip

### DIFF
--- a/src/components/Currency/AvailableConversion.tsx
+++ b/src/components/Currency/AvailableConversion.tsx
@@ -1,11 +1,12 @@
 import { FC, useMemo } from "react";
 import { Chip } from "src/components/Chip/Chip";
-import Tooltip, { TooltipPosition, Variant } from "src/components/Tooltip";
 import { Currency } from "src/types";
 import { CurrencyIcons } from "./CurrencyIcon";
 import { Chips } from "src/components/Chips/Chips";
 import { BudgetCurrencyType, formatMoneyAmount } from "src/utils/money";
 import { useIntl } from "src/hooks/useIntl";
+import Tooltip from "src/components/Tooltip";
+import { TooltipPosition, Variant } from "src/components/Tooltip";
 
 // TODO : doc
 /**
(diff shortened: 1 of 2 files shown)


src/components/Currency/AvailableConversion.tsx
```diff
@@ -129,7 +129,7 @@ export const AvailableConversion: FC<AvailableConversion> = ({
 
   return (
     <>
-      <div {...tooltipIdProps} className="flex flex-row items-center justify-start gap-1">
+      <div {...(currencies ? tooltipIdProps : {})} className="flex flex-row items-center justify-start gap-1">
```
haydencleary, line 132: Maybe it would be cleaner to handle the value of `tooltipIdProps` inside of the `useMemo` above ? Also I think you're missing the `currencies` dependancy in the `useMemo`.

src/components/Currency/AvailableConversion.tsx
```diff
@@ -115,7 +116,10 @@ export const AvailableConversion: FC<AvailableConversion> = ({
     }
 
     return props;
-  }, [currency]);
+  }, [currency, currencies]);
```
haydencleary, line 119: Missing `tooltipId` dependancy too.

src/components/Currency/AvailableConversion.tsx
```diff
@@ -115,7 +116,10 @@ export const AvailableConversion: FC<AvailableConversion> = ({
...
+
+  const tooltipForCurrency = useMemo(() => (currency ? tooltipIdProps : {}), [currency]);
+  const tooltipForCurrencies = useMemo(() => (currencies ? tooltipIdProps : {}), [currencies]);
```
haydencleary, line 122: Can all the tooltip logic be contained in the same `useMemo` ? It's not very clear with 3 `useMemo` hooks ^^'

alexbeno: The problems here are dependent on the style of the component, whether it has an array of currencies (currencies) or just a single currency. That's why I initially implemented this logic in the DOM :/

haydencleary: Ah ok ^^' not easy. Do whatever you think is the simplest !

haydencleary (approved):

PR #1346, opened 2023-10-31
Allow sending rewards using multiple currencies - fixies

alexbeno:
- z-index issue with progress bar when opening selector
- font-weight for currency to heavy on selector (options are ok)
- can we have the whole input selectable when clicking, looks like you cannot click within padding
- cannot enter decimal value, a bit problematic for paying people in ETH
- can we not set 0 as default value but rather show a placeholder that goes away when focusing in field
- can we have menu expanding beyond its selector rather than going to next line when amount is too big
- can we better align the icon on next to network hint when on 2 lines (should be aligned to top)
- progress bar logic is broken, with 67K left to spend after payment, this is not what it should look like
- can we increase font-size for dollar equivalent by 2px (it's too small right now)
- user not appearing as belonging to OD while this guy is already an user (is properly shown upon clicking the name within selector)
- this reward amount has not the proper style
- can we increase a bit the size of the input
- Rewardable items keep being rewardable after they are rewarded (this should only be the case if I cancel reward)


src/components/RewardBudget/RewardBudget.tsx
```diff
@@ -12,11 +12,12 @@ import CheckLine from "src/icons/CheckLine";
 import RewardBudgetBar from "./BudgetBar/RewardBudgetBar";
 import { RewardBudgetUtils } from "./RewardBudget.utils";
 import { Currency } from "src/types";
+import { cn } from "src/utils/cn";
 
 export const RewardBudget: FC<RewardBudgetProps> = props => {
   const { T } = useIntl();
   const [selectedBudget, setSelectedBudget] = useState<WorkEstimationBudgetDetails>(props.budgets[0]);
-  const [amount, setAmount] = useState<number | undefined>(0);
+  const [amount, setAmount] = useState<number | undefined>(undefined);
```
haydencleary, line 20: I think it would be better practice to handle `number | null`
```suggestion
  const [amount, setAmount] = useState<number | null>(null);
```

src/components/RewardBudget/RewardBudget.tsx
```diff
@@ -70,20 +72,27 @@ export const RewardBudget: FC<RewardBudgetProps> = props => {
...
             <FieldInput
               min="0"
+              step="0.01"
+              placeholder={T(`currencies.amount_placeholder.${selectedBudget.currency}`)}
               name="budget-amount-input"
               type="number"
-              value={amount}
-              fieldClassName="flex-1 max-w-[60px] min-w-[60px] w-[60px]"
-              className="h-full flex-1"
+              value={amount === undefined ? "" : amount}
```
haydencleary, line 83: Prefer truthy ternaries
```suggestion
              value={Number.isFinite(amount) ? amount : ""}
```

src/components/RewardBudget/RewardBudget.tsx
```diff
@@ -70,20 +72,27 @@ export const RewardBudget: FC<RewardBudgetProps> = props => {
...
+              fieldClassName="flex-1 max-w-[96px] min-w-[96px] w-[96px]"
```
haydencleary, line 84: Wouldn't this be enough ?
```suggestion
              fieldClassName="flex-1 w-[96px]"
```

PR #1392, opened 2023-11-10
Add disable button on the table fallback for rewards

alexbeno:


src/pages/ProjectDetails/Rewards/List.tsx
```diff
@@ -47,7 +47,10 @@ const RewardList: React.FC = () => {
   });
 
   const rewards = data?.pages.flatMap(page => page.rewards) || [];
-  const isRewardDisabled = projectBudget?.remainingDollarsEquivalent === 0 || rewards.length === 0;
+  const isRewardDisabled =
+    !projectBudget?.remainingDollarsEquivalent ||
+    projectBudget?.remainingDollarsEquivalent === 0 ||
+    rewards.length === 0;
```
haydencleary, line 53: This is the same isn't it ?
```suggestion
  const isRewardDisabled = !projectBudget?.remainingDollarsEquivalent || rewards.length === 0;
```

alexbeno: yes we can do that my bad

alexbeno: fixed !

haydencleary (approved):

PR #1414, opened 2023-11-17
feat: more info optionnal

kaelsky:


src/pages/ProjectDetails/ProjectEdition/EditContext.tsx
```diff
@@ -61,14 +61,14 @@ export const EditContext = createContext<Edit>({
...
+  logoUrl: z.string().optional(),
   inviteGithubUserIdsAsProjectLeads: z.array(z.number()).optional(),
   isLookingForContributors: z.boolean().nullish().optional(),
   longDescription: z.string().min(1),
   moreInfo: z.array(
     z.object({
-      url: z.string().min(1),
-      value: z.string().min(1),
+      url: z.string().min(1).nullish(),
+      value: z.string().min(1).nullish(),
```
haydencleary, line 71: Shouldn't it be ?
```suggestion
      url: z.string().min(1).optional(),
      value: z.string().min(1).optional(),
```

haydencleary: Or is the `min(1)` not compatible with `optional()` ?

kaelsky: Exactly  in this case is the only way to validate correctly  since `nullish` is  explicitly `null` or explicitly `undefined`

haydencleary (approved):

PR #1455, opened 2023-11-23
WIP add <a> instead of window.open

alexbeno:


src/hooks/useContributorProfilePanel/ContributorProfileSidePanel/ReadOnlyView/SocialLink.tsx
```diff
@@ -13,17 +13,23 @@ export default function SocialLink({ link, copyableValue, copyableValueName, tes
...
+  if (link) {
+    return (
+      <a className="rounded-xl bg-white/4" data-testid={testId} href={link} target="_blank" rel="noreferrer">
+        <div className="flex h-10 w-10 flex-row items-center justify-center rounded-xl bg-noise-light text-2xl text-greyscale-200  hover:cursor-pointer hover:opacity-60">
+          {children}
+        </div>
+      </a>
+    );
+  }
+
```
haydencleary, line 25: I think you'd be better of with a polymorphic component.
Something like `const Component = link ? "a" : "div"` then below use it like `<Component ...>`.
Then you could add the necessary props conditionally without repeating the DOM.

alexbeno: i will try :)

haydencleary (approved):

PR #1459, opened 2023-11-24
Rewards Quick actions 

pixelfact:


src/components/TagButton/TagButton.tsx
```diff
@@ -4,7 +4,6 @@ import { MouseEvent } from "react";
 
 export default function TagButton({ children, onClick, ...rest }: ButtonHTMLAttributes<HTMLButtonElement>) {
   const handleClick = (e: MouseEvent<HTMLButtonElement>) => {
-    e.preventDefault();
```
haydencleary, line 7: Why was this removed ? It might be a good idea to leave it to avoid this button submitting any form it may be inside of.

src/pages/Projects/SubmitProject/index.tsx
```diff
@@ -38,6 +38,7 @@ export default function SubmitProject() {
   };
 
   const onClick = () => {
+    console.log("onClick", modalOpened);
```
haydencleary, line 41: Don't forget to remove this ;)

src/types.ts
```diff
@@ -158,6 +158,12 @@ export enum GithubContributionType {
   CodeReview = "CODE_REVIEW",
 }
 
+export enum GithubContributionShortenTypeLabel {
+  Issue = "Issue",
+  PullRequest = "PR",
+  CodeReview = "Code Review",
+}
+
```
haydencleary, line 166: Why is this necessary ? Won't the current type do ?

src/pages/ProjectDetails/Rewards/RewardForm/AutoAdd/QuickActionsMenu.tsx
haydencleary, line 1: Perhaps you could make a generic Menu component that handles the styles and main behaviour then refactor this for your use case. I can imagine other menus coming up in the application.

src/pages/ProjectDetails/Rewards/RewardForm/AutoAdd/QuickActionsMenu.tsx
```diff
@@ -0,0 +1,55 @@
...
+        enterFrom="transform scale-95 opacity-0"
+        enterTo="transform scale-100 opacity-100"
+        leave="transition duration-75 ease-out"
+        leaveFrom="transform scale-100 opacity-100"
+        leaveTo="transform scale-95 opacity-0"
+        className="relative z-10"
+      >
+        <Menu.Items className="absolute -left-6 -top-12 z-40 w-auto min-w-max origin-top-right rounded-xl bg-greyscale-900 p-2 font-walsheim text-sm font-normal shadow-medium outline-none focus:outline-none">
+          <div className="cursor-pointer rounded p-2 hover:bg-white/5">
+            <Menu.Item as="div" onClick={handleAdd}>
```
haydencleary, line 40: Menu items should probably be buttons for accessibility and semantic reasons.

pixelfact: i had a propagation side effet when using a button even with stopPropagation and preventDefault !!

src/components/TagButton/TagButton.tsx
pixelfact: when I pass the button as children of ActionMenu the onclick function is blocked because of the preventDefault

haydencleary (approved):

PR #1486, opened 2023-11-29
E 882 prevent users to click on the more info links when they

alexbeno:


src/pages/ProjectDetails/Overview/OverviewPanel/index.tsx
```diff
@@ -106,10 +112,29 @@ export default function OverviewPanel({
       {moreInfoLink && (
         <Section testId="more-info" icon={SectionIcon.Link} title={T("project.details.overview.moreInfo")}>
           <div data-testid="more-info-link" className="flex overflow-hidden text-sm font-semibold text-spacePurple-500">
-            <ExternalLink
-              text={moreInfoName || moreInfoLink.replace(/^https?:\/\//i, "").replace(/\/$/, "")}
-              url={moreInfoLink}
-            />
+            {isLoggedIn ? (
+              <>
```
kaelsky, line 116: Extra fragment

haydencleary, line 116: Don't need a fragment here

src/pages/ProjectDetails/Overview/OverviewPanel/FakeExternalLink.tsx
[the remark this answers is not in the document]
alexbeno: we don't want to render any link, juste a fake text with icon so it's not the same as the external link components

kaelsky (approved):
👐

PR #1579, opened 2023-12-14
Replace wording

NeoxAzrot:


src/components/RewardBudget/RewardBudget.tsx
```diff
@@ -96,6 +96,8 @@ export const RewardBudget: FC<RewardBudgetProps> = props => {
           >
             {selectedBudget.currency === Currency.USD
               ? T("currencies.network.label_dollar")
+              : selectedBudget.currency === Currency.LORDS
+              ? T("currencies.network.label", { currency: T(`currencies.currency.${Currency.ETH}`) })
               : T("currencies.network.label", { currency: T(`currencies.currency.${selectedBudget.currency}`) })}
```
haydencleary, line 101: Possible to extract this into a function to avoid nested ternaries ?

alexbeno: @NeoxAzrot @haydencleary can we juste replace the translation of network.LORDS by "Ethereum" instead of "Lords" because this network don't exist ?

alexbeno: this will avoid multiple condition for displaying a label ?

alexbeno: `T("currencies.network.label", { currency: T(`currencies.network.${selectedBudget.currency`) })` something like this

alexbeno: same thing here src/App/Stacks/RewardSidePanel/View.tsx

NeoxAzrot: I agree with @alexbeno proprosal

alexbeno (approved):

PR #1580, opened 2023-12-14
Improve layout for contribution tooltip

NeoxAzrot:
e-947-improve-layout-for-contribution-tooltip


src/components/Contribution/ContributionBadge.tsx
```diff
@@ -59,56 +59,58 @@ export function ContributionBadge({
...
-                  />
-                </span>
-                {githubBody ? <p className="line-clamp-2 break-all text-xs text-spaceBlue-200">{githubBody}</p> : null}
-              </div>
-            </div>
-          </div>
-        </Tooltip>
-      ) : null}
-
+  const renderContributionComponent = ({ hasToolTip }: { hasToolTip?: boolean }) => {
```
alexbeno, line 62: this could be a component like ContributionBadge, 

and you can have another component as variant ContributionBadgeWithTooltip that call ContributionBadge. 

Will be more cleaner i think

alexbeno (approved):
good 🔥

PR #1582, opened 2023-12-15
Qa/e 905

haydencleary:


src/components/New/Field/Datepicker.tsx
```diff
@@ -56,19 +82,47 @@ export function Datepicker({ isElevated = false, ...props }: SingleProps | Range
     return Boolean(parseDateString(props.value));
   }, [props.mode, props.value]);
 
+  // If a period is selected use the period's value, otherwise use the custom value
+  const selectedValue = useMemo(() => {
+    return props.selectedPeriod !== Period.Custom && props.periods
```
alexbeno, line 87: if and return for better readability ?

alexbeno: like line 94

src/components/New/Filter/useDatepickerPeriods.ts
alexbeno, line 1: don't know if we want to do it now but maybe we can do 

/Filter/FilterDatepicker/
           FilterDatepicker.tsx
           FilterDatepicker.hooks.ts

src/components/New/Field/Datepicker.tsx
AnthonyBuisset: good catch

src/components/New/Filter/useDatepickerPeriods.ts
haydencleary: d7156548f993f6e201dd40fad770b6918fe55e14

src/components/New/Field/Datepicker.tsx
haydencleary: d7156548f993f6e201dd40fad770b6918fe55e14

alexbeno (approved):
💯

PR #1622, opened 2023-12-26
Add migration folder and middleware.ts to redirect user when env var …

alexbeno:
…is not set


app/opengraph-image.tsx
```diff
@@ -0,0 +1,8 @@
+import Generator from "components/features/seo/image-metadata/commons/generator/generator.tsx";
```
pixelfact, line 1: maybe move to metadata folder ?

pixelfact (commented):
move file if applicable

app/opengraph-image.tsx
alexbeno: this one is a common component that will be for generated all cards inside metadata folder

app/[[...slug]]/layout.tsx
```diff
@@ -0,0 +1,91 @@
...
+import { sharedMetadata } from "../shared-metadata.ts";
+import { ProjectsActions } from "../../actions/Projects/projects.actions.ts";
+import { UsersActions } from "../../actions/Users/users.actions.ts";
+
+type Props = {
+  params: { slug: string[] };
+  searchParams: { [key: string]: string | string[] | undefined };
+};
+
+export async function generateMetadata({ params, searchParams }: Props, parent: ResolvingMetadata): Promise<Metadata> {
```
Comment (pixelfact) [inline app/[[...slug]]/layout.tsx:11]: check if is searchParams is used ?

pixelfact (approved):

PR #1645, opened 2024-01-04
Add iban to individual

alexbeno:


pixelfact (approved):

src/App/Stacks/PayoutInfoSidePanel/components/FormFields/FiatFields.tsx
```diff
@@ -8,9 +8,9 @@ import { RequiredFieldsType } from "src/App/Stacks/PayoutInfoSidePanel/usePayout
```
pixelfact, line 8: Delete BIC_REGEXP import to avoid lint build error

pixelfact:
@alexbeno you need to fix InvoiceSubmission.test.ts

PR #1682, opened 2024-01-15
E-977 - Remove completion score progress bar from navbar

pixelfact:


haydencleary (approved):
Just 1 non-blocking comment

PR #1690, opened 2024-01-17
B 1312 posthog identify users and track log in log out

haydencleary:


haydencleary:
> Just missing the `created_at` but you can check the rest @pixelfact

Done

pixelfact (approved):

PR #1700, opened 2024-01-22
E-1014 : Ask users to create invoice in dollar equivalent

pixelfact:


src/_pages/Rewards/InvoiceSubmission/View.tsx
```diff
@@ -129,17 +129,10 @@ export function buildHiddenFields({
...
-      : formatList([
-          `IBAN: ${payoutInfo?.payoutSettings?.sepaAccount?.iban}`,
-          `BIC: ${payoutInfo?.payoutSettings?.sepaAccount?.bic}`,
-        ]),
+    payout_info: "",
     total_amount: formatMoneyAmount({
-      amount: paymentRequests.map(p => p.amount.total).reduce((acc, amount) => acc + amount, 0),
-      currency: paymentRequests.at(0)?.amount.currency,
+      amount: paymentRequests.map(p => p.amount.dollarsEquivalent ?? 0).reduce((acc, amount) => acc + amount, 0),
+      currency: "USD",
```
haydencleary, line 135: Haven't we got a constant for this value?

haydencleary (approved):

src/_pages/Rewards/InvoiceSubmission/View.tsx
```diff
@@ -118,7 +119,9 @@ export function buildHiddenFields({
           `#${pretty(p.id)} - ${formatDate(new Date(p.requestedAt))} (${formatMoneyAmount({
             amount: p.amount.total,
             currency: p.amount.currency,
-          })})`
+          })} ${
+            p.amount.currency !== Currency.USD ? "~ " + p.amount.dollarsEquivalent?.toFixed(5) + " " + Currency.USD : ""
```
alexbeno, line 123: you should handle the fact that STARK don't have dollar equivalent

alexbeno (changes requested):
Handle STRK no dollar equivalent

src/_pages/Rewards/InvoiceSubmission/View.tsx
haydencleary: oops

pixelfact: STRK and OP since locked should been excluded from the Pending invoice list

alexbeno: ok so they are no way that dollar equivalent was write as 0 for STARK ?

alexbeno: don't know if we want to be bullet proof to this and add a little condition in case of bug on the locked status

alexbeno (approved):

PR #1702, opened 2024-01-22
e-1019-handle-the-fact-a-lead-may-have-connected-only-some

NeoxAzrot:
e-1019-handle-the-fact-a-lead-may-have-connected-only-some


components/features/add-missing-repositories/add-missing-repositories.tsx
```diff
@@ -0,0 +1,44 @@
...
+          </Typography>
+
+          <Typography variant="body-xs">
+            <Translate token="callout.missingRepositories.description" />
+          </Typography>
+        </Flex>
+      </Flex>
+
+      <a href={url} target="_blank" rel="noreferrer">
+        <Button
```
haydencleary, line 31: Maybe give `Button` an `asChild` prop like in [radix-ui](https://www.radix-ui.com/themes/docs/components/button). That way the anchor would _be_ the button, instead of wrapping it.

haydencleary (approved):

PR #1709, opened 2024-01-23
e-1016-update-user-menu-and-move-feedback-button-in-it

NeoxAzrot:
e-1016-update-user-menu-and-move-feedback-button-in-it


src/App/Layout/Header/ProfileButton/View.tsx
```diff
@@ -4,22 +4,52 @@ import { useStackContributorProfile, useStackPayoutInfo } from "src/App/Stacks/S
...
+      "cursor-pointer ui-active:bg-white/4": !disabled,
+      "cursor-default": disabled,
+      "text-greyscale-50": !secondary,
+      "text-spaceBlue-200": secondary,
+    })}
+    onClick={onClick}
+  >
+    {children}
+  </Menu.Item>
+);
```
haydencleary, line 34: Perhaps extract into its own file ?

src/App/Layout/Header/ProfileButton/LogoutButton.tsx
haydencleary, line 1: Now that the way the logout button is used has changed, it might make more sense to put the click handler into a custom hook and just use that.

src/App/Layout/Header/ProfileButton/index.tsx
```diff
@@ -16,15 +20,25 @@ const ProfileButton = () => {
...
     login,
     isMissingPayoutSettingsInfo: payoutSettingsInvalid && !onboardingInProgress,
     hideProfileItems: onboardingInProgress,
+    openFeedback,
   };
 
-  return isXl ? <View {...props} /> : <ViewMobile {...props} />;
+  return (
+    <>
+      <FeedbackButton customButton={<Fragment />} ref={feedbackButtonRef} />
```
haydencleary, line 38: Not sure about this `customButton` prop, is this necessary ?

src/App/Layout/Header/ProfileButton/LogoutButton.tsx
NeoxAzrot: I tried, but the file has to be a Reac Component. How do you imagine doing it?

src/App/Layout/Header/ProfileButton/index.tsx
NeoxAzrot: This allows you to define it at parent level without displaying it. Because in in the menu, we refresh states on click, so the feedback panel never shows

src/App/Layout/Header/ProfileButton/View.tsx
haydencleary: Ok thats fine 👍

haydencleary (approved):

PR #1726, opened 2024-01-26
E 1043 uniformise links style across product

NeoxAzrot:
e-1043-uniformise-links-style-across-product


components/ds/link/link.tsx
```diff
@@ -0,0 +1,46 @@
...
+    </>
+  );
+}
+
+// TODO: Change to next/link after the migration
+export function Link({ href, target, rel, onClick, children, className, ...props }: TLink.Props) {
+  const isExternal = href?.toString().startsWith("http") ?? false;
+
+  const targetProps: HTMLAttributeAnchorTarget | undefined = target ? target : isExternal ? "_blank" : undefined;
+  const relProps = rel ? rel : isExternal ? "noopener noreferrer" : undefined;
```
alexbeno, line 23: `
const relExternal = isExternal ? "noopener noreferrer" : undefined;
const relProps = rel || relExternal;
`

components/ds/link/link.variants.ts
```diff
@@ -0,0 +1,14 @@
+import { tv } from "tailwind-variants";
+
+export const linkVariants = tv({
+  base: "group/link inline-flex items-center gap-1 truncate hover:text-spacePurple-300",
```
alexbeno, line 4: you don't use `group/link` inside your code ?

src/App/Stacks/ContributorProfileSidePanel/ReadOnlyView/IntroSection.tsx
```diff
@@ -154,37 +151,37 @@ export default function IntroSection({ isOwn, isPublic, profile, setEditMode }:
...
         {profile.htmlUrl && (
           <SocialLink testId="github" link={profile.htmlUrl}>
-            <GithubLogo />
+            <Icon remixName="ri-github-fill" size={24} />
           </SocialLink>
         )}
         {telegram && (
           <SocialLink testId="telegram" link={telegram}>
-            <Telegram className="fill-greyscale-200" size={20} />
+            <Telegram size={24} />
```
alexbeno, line 159: why did you remove the className here ?

src/App/Stacks/ContributorProfileSidePanel/ReadOnlyView/SocialLink.tsx
```diff
@@ -13,7 +14,7 @@ export default function SocialLink({ link, copyableValue, copyableValueName, tes
   const showToaster = useShowToaster();
   const { T } = useIntl();
 
-  const Component = link ? "a" : "div";
+  const Component = link ? "a" : "button";
```
alexbeno, line 17: "div" was not used ?

alexbeno (changes requested):

src/components/Project/Overview/OverviewSponsors.tsx
```diff
@@ -19,9 +20,18 @@ export const ProjectOverviewSponsors = ({ sponsors }: ProjectOverviewSponsorsPro
...
+            {sponsor.url ? (
+              <Link href={sponsor.url} className="flex items-center gap-2">
+                <RoundedImage alt={sponsor.name} rounding={Rounding.Circle} size={ImageSize.Sm} src={sponsor.logoUrl} />
+                {sponsor.name}
+              </Link>
+            ) : (
+              <Typography variant="body-s" className="flex items-center gap-2">
+                <RoundedImage alt={sponsor.name} rounding={Rounding.Circle} size={ImageSize.Sm} src={sponsor.logoUrl} />
+                {sponsor.name}
+              </Typography>
```
haydencleary, line 33: You could store the rounded image in a variable to avoid repeating it

src/components/Project/Overview/OverviewRepos/GithubRepoDetails/GithubLink/index.tsx
```diff
@@ -1,18 +1,16 @@
...
   return (
-    <a
-      href={link}
-      target="_blank"
-      className="flex h-8 w-8 justify-center rounded-xl border border-neutral-100 bg-white/5 grayscale hover:cursor-pointer hover:opacity-60"
-      rel="noreferrer"
-    >
-      <img className="w-3.5 fill-neutral-100" alt="GitHub" src={IMAGES.svg.logos.github} loading="lazy" />
+    <a href={link} target="_blank" rel="noreferrer">
+      <Button size="s" variant="secondary" iconOnly>
```
haydencleary, line 11: Not a fan of this, tried using the `as` prop ?

src/components/Contributor/index.tsx
```diff
@@ -51,6 +33,29 @@ export default function Contributor({ contributor, clickable, className }: Props
...
+export default function Contributor({ contributor, clickable, className }: Props) {
+  const [open] = useStackContributorProfile();
+
+  const handleClick = (e: SyntheticEvent) => {
+    e.preventDefault();
+    open({ githubUserId: contributor.githubUserId });
+  };
+
+  return (
+    <>
```
haydencleary, line 49: Don't need the fragment here if you return the ternary

components/ds/link/link.tsx
```diff
@@ -0,0 +1,46 @@
...
+      {children}
+
+      {isExternal ? <Icon remixName="ri-external-link-line" className="invisible group-hover/link:visible" /> : null}
+    </>
+  );
+}
+
+// TODO: Change to next/link after the migration
+export function Link({ href, target, rel, onClick, children, className, ...props }: TLink.Props) {
+  const isExternal = href?.toString().startsWith("http") ?? false;
```
haydencleary, line 20: I think this should be passed explicitly as a prop

components/ds/link/link.tsx
haydencleary, line 23: You could have a base `rel` value and extend it with the `isExternal` value and the `rel` prop. It would avoid nested ternaries

components/ds/link/link.tsx
```diff
@@ -0,0 +1,46 @@
...
+
+  const targetProps: HTMLAttributeAnchorTarget | undefined = target ? target : isExternal ? "_blank" : undefined;
```
haydencleary, line 22: Same as `rel`

components/ds/link/link.tsx
```diff
@@ -0,0 +1,46 @@
...
+          <LinkContent isExternal={isExternal}>{children}</LinkContent>
+        </NavLink>
+      ) : null}
+
+      {onClick ? (
+        <span onClick={onClick} className={cn(linkVariants({ ...props }), "cursor-pointer", className)}>
+          <LinkContent isExternal={isExternal}>{children}</LinkContent>
+        </span>
+      ) : null}
+    </>
```
haydencleary, line 44: It might be simpler to split these into separate components. In the case of an `onClick` I would use a `<button>` instead of a `<span>`

components/ds/link/link.tsx
NeoxAzrot: Rel use only if external, and if external use blank or take target

components/ds/link/link.tsx
NeoxAzrot: Do 2 returns instead, with one return null

src/components/Project/Overview/OverviewSponsors.tsx
NeoxAzrot: Do a variable which return both

components/ds/link/link.variants.ts
NeoxAzrot: I do, with external

src/App/Stacks/ContributorProfileSidePanel/ReadOnlyView/IntroSection.tsx
NeoxAzrot: It's done with currentColor now

src/App/Stacks/ContributorProfileSidePanel/ReadOnlyView/SocialLink.tsx
NeoxAzrot: We want the component to be a button if there is an onClick action

alexbeno (approved):

PR #1744, opened 2024-01-30
E 1061 build new account page for users move existing sections

NeoxAzrot:
e-1061-build-new-account-page-for-users-move-existing-sections


app/migration/settings/components/settings-header/settings-header.tsx
```diff
@@ -7,15 +7,17 @@ import { Typography } from "components/layout/typography/typography";
 
 export function SettingsHeader({ icon, title, subtitle, children }: TSettingsHeader.Props) {
   return (
-    <Card background="base">
+    <Card background="base" className="mb-2">
```
haydencleary, line 10: The parent should be responsible for the spacing of it's children. Ex: gap with flex/grid

app/migration/settings/components/sidebar.tsx
```diff
@@ -6,39 +6,54 @@ import { usePathname } from "next/navigation";
...
     <LayoutSidebar
-      // TODO
+      // TODO: mobile header
       mobileHeader={<div>Mobile header</div>}
     >
       {({ closePanel }) => (
-        <div className="flex w-full flex-col gap-4 divide-neutral-700 xl:gap-6 xl:divide-y">
-          <div>{/* TODO user card */}</div>
+        <div className="flex w-full flex-col gap-4 xl:gap-6">
+          <Flex
```
haydencleary, line 41: If there is no user shouldn't this be a skeleton ?

app/migration/settings/profile/features/form/form.utils.ts
haydencleary, line 1: I think it would be a good idea to add unit tests for these utils

app/migration/settings/profile/features/form/informations/informations.tsx
```diff
@@ -0,0 +1,120 @@
...
+
+  const {
+    mutate: uploadProjectLogo,
+    isPending: loadingUploadLogo,
+    isSuccess: successUploadLogo,
+  } = MeApi.mutations.useUploadProfilePicture({
+    options: {
+      onSuccess: data => {
+        setValue("avatarUrl", data.url);
+      },
```
haydencleary, line 27: No error handling

components/ds/form/input/input.tsx
```diff
@@ -0,0 +1,24 @@
...
+        innerWrapper: "gap-2",
+        input: "!p-0 font-walsheim text-sm leading-none",
+      }}
+      variant="bordered"
+      labelPlacement="outside"
+      type={type}
+      value={value}
+      onChange={onChange}
+      startContent={startContent}
+      endContent={endContent}
```
haydencleary, line 20: Don't need to extract these from the props above, they could stay inside `...props`

components/ds/form/textarea/textarea.tsx
```diff
@@ -0,0 +1,24 @@
...
+        innerWrapper: "gap-2",
+        input: "!p-0 font-walsheim text-sm leading-none",
+      }}
+      variant="bordered"
+      labelPlacement="outside"
+      type={type}
+      value={value}
+      onChange={onChange}
+      startContent={startContent}
+      endContent={endContent}
```
haydencleary, line 20: Same as input

src/App/Stacks/ContributorProfileSidePanel/ReadOnlyView/IntroSection.tsx
```diff
@@ -53,7 +53,7 @@ export default function IntroSection({ isOwn, isPublic, profile, setEditMode }:
     return find.contact;
   };
 
-  const email = findContact("EMAIL");
+  // const email = findContact("EMAIL");
```
haydencleary, line 56: If we're not inputting the email, might as well remove commented code concerning the email address

app/migration/settings/components/settings-header/settings-header.tsx
NeoxAzrot: This is specific, because the gap is 16px for each card, and only the header has a gap of 24px.
So, I let gap-4 on the parent for the others and just add an additional margin-bottom for the header, isn't it working?

app/migration/settings/components/sidebar.tsx
NeoxAzrot: How it could have no user, it's the settings menu

components/ds/form/textarea/textarea.tsx
NeoxAzrot: I could merge both with a props to return Textarea or Input?

haydencleary (approved):

PR #1747, opened 2024-01-30
Feat/column template

haydencleary:
Started working on new Settings layout with Sami. The goal is to have basic components we can use later to rebuild the sidebar/column layout used for projects and now user settings.

The new pages are not protected as there are no links to them right now, what do you think @alexbeno @pixelfact  ?

### DIFF
--- a/app/settings/components/sidebar.tsx
+++ b/app/settings/components/sidebar.tsx
@@ -1,0 +1,67 @@
+"use client";
+
+import { useAuth0 } from "@auth0/auth0-react";
+import Link from "next/link";
+import { usePathname } from "next/navigation";
+
+import GithubLink, { Variant as GithubLinkVariant } from "src/App/Layout/Header/GithubLink";
+import { useIntl } from "src/hooks/useIntl";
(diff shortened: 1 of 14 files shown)


alexbeno:
@haydencleary maybe you can put your page in de the migration folder !

app/settings/components/sidebar.tsx
```diff
@@ -0,0 +1,67 @@
...
+                key={tab.path}
+                href={tab.path}
+                onClick={closePanel}
+                className={cn("rounded-xl px-4 py-2.5 text-base", {
+                  "bg-white/8 text-white": pathname === tab.path,
+                  "text-neutral-400": pathname !== tab.path,
+                })}
+              >
+                {tab.label}
+              </Link>
```
alexbeno, line 54: maybe we should have a SideBarItem component ?

components/layout/background/background.tsx
alexbeno, line 1: components/layout/pages/space-background already exist in the new components

alexbeno (approved):
approved

PR #1780, opened 2024-02-07
styles: tooltip arrow color

haydencleary:


components/ds/tooltip/tooltip.tsx
```diff
@@ -8,6 +8,10 @@ export function Tooltip({ as: Component = "div", children, ...props }: TTooltip.
       {...props}
       showArrow={true}
       classNames={{
+        base: [
+          // arrow color
+          "before:bg-greyscale-800",
+        ],
         content: ["p-4 bg-greyscale-800", "od-text-body-s"],
```
NeoxAzrot, line 15: Why in array?

pixelfact (approved):

components/ds/tooltip/tooltip.tsx
haydencleary: a7c18044d570b77eeb35e05050b3a27fc34df765

PR #1785, opened 2024-02-07
Feat/avatar component

haydencleary:
First step towards new `User` component


components/ds/avatar/avatar.stories.tsx
```diff
@@ -0,0 +1,60 @@
...
+      </div>
+    );
+  },
+};
+
+export const Radius: Story = {
+  render: args => {
+    return (
+      <div className="flex w-full items-center gap-2">
+        <Avatar {...defaultProps} {...args} radius="full" />
```
alexbeno, line 50: Can you add the same variant as the thumbnail component ? the size and variant as the same as figma ?

alexbeno (approved):
Nice one

components/ds/avatar-group/avatar-group.tsx
```diff
@@ -0,0 +1,20 @@
...
+    <div className={cn(avatarGroupVariants({ ...props, size: avatarProps?.size }), className)}>
+      {avatars.map((avatar, index) => (
+        <Avatar
+          key={index}
+          {...avatar}
+          {...avatarProps}
+          className="transition-transform hover:-translate-x-1/4 hover:last:translate-x-0"
+        />
+      ))}
+    </div>
```
alexbeno, line 18: Why did you not use this : 

https://nextui.org/docs/components/avatar#avatar-group

haydencleary: I did at first but it made things more complicated than necessary. The only thing I found it useful for was the animation so I did it myself.

haydencleary:
Just realised a component I deleted was still in use, going to fix this now.

haydencleary:
> Just realised a component I deleted was still in use, going to fix this now.

Fixed

components/ds/avatar-group/avatar-group.tsx
alexbeno: ok in this case can you add a props to select how many avatar we want to show and if exceed we add a +N ( like the AvailableConversion component ? ) or we keep this for later ?

haydencleary: We can handle this when we get there

PR #1802, opened 2024-02-09
E 1091 allow users to trigger a refresh of their github account

NeoxAzrot:
e-1091-allow-users-to-trigger-a-refresh-of-their-github-account


haydencleary (approved):
Just one non-blocking remark

PR #1804, opened 2024-02-09
E 1059 change look and feel for usd

haydencleary:


alexbeno (changes requested):
i think you have to go on some formatMoneyAmount ( on the news features ) that already put showCurrency to false and manually add USD at the end of the string ?

alexbeno (approved):

PR #1834, opened 2024-02-15
Add error catching for stack panels

alexbeno:


src/libs/react-stack/context/stack.context.provider.tsx
```diff
@@ -91,21 +101,31 @@ export default function ReactStackprovider({ children }: reactStackContextProps)
...
+    try {
+      if (!stacks.state[stack.state.name]) {
+        setStacks(prev => ({
+          ...prev,
+          [stack.state.name]: stack,
+        }));
+      }
+    } catch (e) {
+      console.warn("stacks - registerStack error", e);
+      // registerStack error
```
haydencleary, line 113: Can remove these comments

pixelfact (approved):

src/libs/react-stack/context/stack.context.provider.tsx
```diff
@@ -333,8 +391,16 @@ export default function ReactStackprovider({ children }: reactStackContextProps)
...
   const onCloseByPanelId = useCallback(
     (name: string, panelId: string) => {
-      const { panel, stack } = getPanelFromStackNameAndPanelId(name, panelId);
-      removePanel(panel, stack);
+      try {
+        const getPanel = getPanelFromStackNameAndPanelId(name, panelId);
+        if (getPanel) {
+          const { panel, stack } = getPanel;
+          removePanel(panel, stack);
+        }
```
haydencleary, line 399: Isn't there an error in the else case ?

alexbeno: i prefere catch all function in case of a similare side effect in the future

haydencleary (approved):

PR #1850, opened 2024-02-16
feat: rewards

haydencleary:


app/rewards/page.tsx
```diff
@@ -0,0 +1,5 @@
+import LegacyRewards from "src/_pages/Rewards";
+
+export default function Rewards() {
```
alexbeno, line 3: To keep the same naming `RewardsPage`

PR #1857, opened 2024-02-19
Clean up contributions, rewards and settings auth guard

haydencleary:


app/rewards/page.tsx
```diff
@@ -1,5 +1,11 @@
...
+import { withAuthenticationRequired } from "@auth0/auth0-react";
+
 import LegacyRewards from "src/_pages/Rewards";
 
-export default function RewardsPage() {
+function RewardsPage() {
   return <LegacyRewards />;
 }
+
+export default withAuthenticationRequired(RewardsPage);
```
alexbeno, line 11: maybe we want to put this into the layout ? as all the page of settings should be with auth ?

src/App/Stacks/RewardSidePanel/View.tsx
```diff
@@ -1,11 +1,10 @@
 import { UseMutateFunction } from "@tanstack/react-query";
 import { PropsWithChildren, useMemo, useState } from "react";
-import { NavLink, matchPath, useLocation } from "react-router-dom";
+import { NavLink } from "react-router-dom";
```
alexbeno, line 3: don't use baseLink ?

app/rewards/page.tsx
haydencleary: Doesn't work as the layout is server rendered and I think we should use the server as much as possible

PR #1858, opened 2024-02-19
feat: public profile

haydencleary:


constants/router.ts
```diff
@@ -27,6 +27,7 @@ export const NEXT_ROUTER = {
     billing: "/settings/billing",
   },
   publicProfile: {
+    base: "/u",
```
alexbeno, line 30: why do you need base ?

PR #1859, opened 2024-02-19
Ft next router/project edition

alexbeno:


src/_pages/ProjectDetails/ProjectEdition/components/ConfirmationModal/ConfirmationModal.tsx
```diff
@@ -1,35 +1,44 @@
...
+  const [, setBlockState] = block.state;
+  const [showConfirmation] = block.confirmation;
+
+  useEffect(() => {
+    const shouldBlock = !githubWorklow.inGithubWorkflow && !!form?.formState.isDirty;
+    if (shouldBlock) {
+      setBlockState(true);
+    } else {
+      setBlockState(false);
+    }
```
haydencleary, line 23:
```suggestion
      setBlockState(shouldBlock);
```

providers/navigation-state/navigation-state.tsx
```diff
@@ -0,0 +1,67 @@
...
+    if (!value) {
+      onCancel();
+    } else {
+      setShouldBlockNavigation(true);
+    }
+  }
+
+  const block = useMemo(
+    () => ({
+      should: shouldBlockNavigation,
```
haydencleary, line 42: Don't need `should` the state getter is enough

src/_pages/ProjectDetails/ProjectEdition/components/ConfirmationModal/ConfirmationModal.tsx
alexbeno: if true return true

haydencleary (approved):

PR #1889, opened 2024-02-22
feat: Table component

haydencleary:


components/ds/table/table.tsx
alexbeno, line 1: Can we add that ability to add classNames unitary for row and headers ?

alexbeno (approved):

PR #1894, opened 2024-02-23
Migrate billing profile into new dedicated tab

alexbeno:


app/migration/settings/billing/[id]/general-information/features/profile/profile-company/profile-company.tsx
```diff
@@ -0,0 +1,59 @@
...
+export function ProfileCompany({ profile }: TProfileCompany.Props) {
+  const registrationDate = useMemo(() => {
+    if (profile.registrationDate) {
+      return format(new Date(profile.registrationDate), "MMM dd, yyyy");
+    }
+
+    return profile.registrationDate;
+  }, [profile]);
+
+  const isYesOrNO = (yes: ReactNode, no: ReactNode, value?: boolean) => {
```
Comment (pixelfact) [inline app/migration/settings/billing/[id]/general-information/features/profile/profile-company/profile-company.tsx:19]: Maybe move this logic into a small dedicated component that take this same values as props ?
And use it in both profile types company + individual 

Also you can shorten the condition inside `if (value)`

Comment (alexbeno) [reply app/migration/settings/billing/[id]/general-information/features/profile/profile-company/profile-company.tsx:19]: good point !

pixelfact (commented):
non-blocking improvement proposal

PR #1919, opened 2024-03-04
Rewards included in invoices cannot be canceled

pixelfact:


src/translations/en.json
```diff
@@ -661,7 +661,8 @@
             "cancelButton": "Go back",
             "description": "Once cancelled, you won't be able to restore this reward",
             "title": "Are you sure?"
-          }
+          },
+          "tooltip": "Rewards in processing status cannot be cancelled."
```
haydencleary, line 665:
```suggestion
          "tooltip": "Rewards that are being processed cannot be cancelled."
```

haydencleary (approved):

PR #1939, opened 2024-03-18
Delete billing profile

pixelfact:


app/settings/billing/[id]/general-information/page.tsx
```diff
@@ -34,23 +36,26 @@ function SettingsBillingPage() {
     }
   }, [open, isPanelHasOpenedState]);
 
+  const renderValue = useMemo(() => {
```
Comment (haydencleary) [inline app/settings/billing/[id]/general-information/page.tsx:39]: Perhaps return `null` if `profile` is undefined ?

haydencleary (approved):
Just one change

app/settings/billing/[id]/general-information/page.tsx
Comment (pixelfact) [reply app/settings/billing/[id]/general-information/page.tsx:39]: you can find in next line the return null if !profile, it's not in the memo because we need to return null instead of the component

PR #1960, opened 2024-03-20
handle enable bp icon

pixelfact:


app/settings/billing/[id]/general-information/features/manage-billing-profile/manage-billing-profile.tsx
```diff
@@ -81,10 +81,21 @@ export function ManageBillingProfile({ actionType }: TManageBillingProfile.Props
     }
   }
 
+  const iconName = useMemo(() => {
```
Comment (haydencleary) [inline app/settings/billing/[id]/general-information/features/manage-billing-profile/manage-billing-profile.tsx:84]: Could just be a static map defined outside of the component

haydencleary (approved):

PR #2004, opened 2024-04-16
Hackathons single ( V1 )

alexbeno:


components/features/seo/image-metadata/commons/content/content.tsx
```diff
@@ -22,6 +22,16 @@ export function ImageMetadataContent({ title, description }: TContent.Props) {
       >
         {title}
       </div>
+      {!!subtitle && (
```
pixelfact, line 25: Same as above

pixelfact (approved):
check comments, not mandatory

PR #2005, opened 2024-04-17
revamp insights component

pixelfact:


haydencleary (approved):
You could use the new table instead, but that could be refactored later.

PR #2041, opened 2024-04-25
Move use Intl Hook

NeoxAzrot:
fix/use-intl


translations/v2/index.ts
haydencleary, line 1: Is the V2 folder necessary ? As you have merged everything I don't think we need to specify this.

haydencleary (approved):

PR #2046, opened 2024-04-25
Open Request payment on click on status

alexbeno:


src/App/Stacks/RewardSidePanel/View.tsx
```diff
@@ -99,6 +99,8 @@ export default function View({
         status={data.status}
         dates={{ unlockDate: data?.unlockDate, processedAt: data?.processedAt }}
         billingProfileId={data?.billingProfileId}
+        rewardId={data?.id}
+        shouldOpenRequestPayment={true}
```
pixelfact, line 103: do we need to add this prop knowing that we always want to trigger the payment flow on the "request payment" status ?

pixelfact (commented):
check comment

haydencleary (approved):

PR #2052, opened 2024-04-26
Create profile card

pixelfact:


translations/v2/en/features/profile-card.json
```diff
@@ -0,0 +1,11 @@
+{
+  "position": "__position__th",
```
haydencleary, line 2: This isn't universel, come and see me and I'll try to explain ^^

components/features/profile-card/profile-card.tsx
```diff
@@ -0,0 +1,107 @@
...
+        </Typography>
+      );
+    }
+    return (
+      <Typography
+        variant="body-s"
+        className="line-clamp-2 italic text-spaceBlue-100"
+        translate={{ token: "v2.features.profileCard.emptyBio" }}
+      />
+    );
```
haydencleary, line 50: Is it necessary to show an empty message ?

components/features/profile-card/profile-card.tsx
```diff
@@ -0,0 +1,107 @@
...
+            <ProfileStatItem
+              icon="ri-stack-line"
+              token="v2.features.profileCard.counters.contributionCount"
+              count={contributionCount}
+            />
+            <ProfileStatItem
+              icon="ri-medal-2-fill"
+              token="v2.features.profileCard.counters.rewardCount"
+              count={rewardCount}
+            />
```
haydencleary, line 85: Is it worth putting this in an external component ? Could just loop over an array of data.

components/features/profile-card/profile-card.tsx
```diff
@@ -0,0 +1,107 @@
...
+            params={{ count: contributedProjectCount }}
+          />
+        </Tag>
+        <Tag size="medium">
+          <Icon remixName="ri-star-line" size={16} />
+          <Translate
+            token="v2.features.profileCard.counters.leadedProjectCount"
+            params={{ count: leadedProjectCount }}
+          />
+        </Tag>
```
haydencleary, line 103: Same as profile stat item, could loop over an array ? 🤷‍♂️

haydencleary (changes requested):

utils/profile/ordinal-position-suffix.ts
```diff
@@ -0,0 +1,19 @@
+export function getOrdinalSuffix(position: number): string {
+  // Check the last two digits to handle special cases like 11th, 12th, 13th, etc.
+  const lastTwoDigits = position % 100;
+  if (lastTwoDigits >= 11 && lastTwoDigits <= 13) {
+    return `${position}th`;
+  }
```
haydencleary, line 6: I think this could work for >= 10 and <=19, might as well do that work here. Other than that looks good 👍

components/features/profile-card/profile-card.tsx
pixelfact: not anymore

translations/v2/en/features/profile-card.json
pixelfact: deleted

haydencleary (approved):

PR #2061, opened 2024-04-29
B 1537 build the technoecosystem accordion

pixelfact:


components/ds/avatar/avatar.variants.ts
```diff
@@ -56,6 +58,11 @@ export const avatarVariants = tv({
       size: "3xl",
       class: "h-20 w-20 rounded-2xl border-[6px]",
     },
+    {
+      shape: "rectangle",
+      size: "rectangleSize",
```
haydencleary, line 63: Not sure about these new values. The `shape` prop was designed to handle the radius of the avatar. Also, to me an avatar suggests that the width and height will be the same. Maybe your case should be a one-off ?

app/migration/u/[githubLogin]/features/details-accordion/details-accordion.tsx
```diff
@@ -0,0 +1,42 @@
+"use client";
```
Comment (haydencleary) [inline app/migration/u/[githubLogin]/features/details-accordion/details-accordion.tsx:1]: Is it required to be a client component ? No hooks or events here

haydencleary (changes requested):

app/migration/u/[githubLogin]/features/details-accordion/details-accordion.tsx
Comment (pixelfact) [reply app/migration/u/[githubLogin]/features/details-accordion/details-accordion.tsx:1]: Yes since the accordion from NextUi wrap in it some state handling ?! it was server at first then i got error...

components/ds/avatar/avatar.variants.ts
pixelfact: yes I wasn't really sure if it could be considered an avatar variant. We discuss it live to find a better choice ?

components/ds/avatar/avatar.variants.ts
```diff
@@ -4,6 +4,7 @@ export const avatarVariants = tv({
   base: "relative box-content flex flex-shrink-0 items-center justify-center overflow-hidden border-greyscale-50/12 bg-greyscale-50/8",
   variants: {
     shape: {
+      rectangle: "",
```
haydencleary, line 7: I think you can remove this

haydencleary (changes requested):

haydencleary (approved):

PR #2064, opened 2024-04-30
Add badge in menu and rewards count

alexbeno:


src/App/Layout/Header/View.tsx
```diff
@@ -35,11 +37,28 @@ export default function HeaderView({ menuItems, impersonating = false }: HeaderV
...
   const hideHeader = (isMatchProjectDetail || isMatchSettings) && !isXl;
 
+  const rewardSum = useMemo(
+    () => billingProfile?.billingProfiles?.reduce((acc, profile) => acc + profile.invoiceableRewardCount, 0),
+    [billingProfile]
+  );
+
+  const rewardBadgeContent = useMemo(() => {
+    if (!rewardSum) {
+      return undefined;
```
haydencleary, line 52:
```suggestion
      return null;
```

src/App/Layout/Header/View.tsx
```diff
@@ -35,11 +37,28 @@ export default function HeaderView({ menuItems, impersonating = false }: HeaderV
...
+    }
+
+    if (rewardSum > 9) {
+      return `+9`;
```
haydencleary, line 56:
```suggestion
      return "+9";
```

src/App/Layout/Header/View.tsx
```diff
@@ -35,11 +37,28 @@ export default function HeaderView({ menuItems, impersonating = false }: HeaderV
...
+    }
+
+    return `${rewardSum}`;
```
haydencleary, line 59:
```suggestion
    return rewardSum;
```

haydencleary (changes requested):

src/App/Layout/Header/View.tsx
alexbeno: should be undefined for the next ui component

src/App/Layout/Header/View.tsx
alexbeno: content should be a string

haydencleary (approved):

PR #2176, opened 2024-05-24
Disable form if not open for votes

haydencleary:


app/c/[committeeId]/jury/utils/status-context.ts
```diff
@@ -0,0 +1,4 @@
+import { MyCommitteeAssignmentsResponse } from "api-client/resources/me/types";
+import { createContext } from "react";
+
+export const StatusContext = createContext<MyCommitteeAssignmentsResponse["status"] | "">("");
```
Comment (alexbeno) [inline app/c/[committeeId]/jury/utils/status-context.ts:4]: you don't want to return an objet instead of single value, thank to that we can extends the without refactoring status implementation as the features growth

alexbeno (approved):

PR #2187, opened 2024-05-28
Update open graph card on public profile

NeoxAzrot:
e-1410-update-card-with-ranking

### DIFF
--- a/components/features/seo/image-metadata/public-profile/components/content/content-user.tsx
+++ b/components/features/seo/image-metadata/public-profile/components/content/content-user.tsx
@@ -1,11 +1,14 @@
 import process from "process";
+import { getOrdinalSuffix } from "utils/profile/ordinal-position-suffix";
 
 interface Props {
   login: string;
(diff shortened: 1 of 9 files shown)


components/features/seo/image-metadata/public-profile/components/content/content-user.tsx
```diff
@@ -35,8 +38,8 @@ export function ContentUser({ login, title, image }: Props) {
       <img
         src={optimizeSrc}
         alt="user-image"
-        width="108"
-        height="108"
+        width="196"
+        height="196"
```
haydencleary, line 42: This seems like a big size jump to me, done on purpose ?

NeoxAzrot: Yes : "We need to increase the size of the picture so that its the same height as the name + tagline + ranking"

haydencleary (approved):

PR #2327, opened 2024-06-13
B 1945 paper

haydencleary:


components/atoms/typo/typo.loading.tsx
```diff
@@ -1,3 +1,3 @@
-export const TypoLoading = () => {
+export function TypoLoading() {
   return <div />;
```
pixelfact, line 2: We should return a Skeleton here ?!

pixelfact (approved):

PR #2332, opened 2024-06-14
B 1952 avatar

alexbeno:


haydencleary:
Just for an avatar maybe NextUI isn't required ?

components/atoms/avatar/avatar.loading.tsx
```diff
@@ -0,0 +1,11 @@
+import { cn } from "@nextui-org/react";
+
+import { TAvatarProps } from "components/atoms/avatar/avatar.types";
+import { getAvatarImageSize } from "components/atoms/avatar/avatar.utils";
+import { SkeletonEl } from "components/ds/skeleton/skeleton";
+
+export function AvatarLoading({ size, shape }: TAvatarProps) {
+  const [w, h] = getAvatarImageSize(size);
+
+  return <SkeletonEl variant="rounded" width={w} height={h} className={cn({ "rounded-full": shape !== "square" })} />;
```
haydencleary, line 10: Don't we need to apply the correct radius depending on the avatar size ?

components/atoms/avatar/avatar.stories.tsx
```diff
@@ -0,0 +1,126 @@
...
+import { Avatar } from "./variants/avatar-default";
+
+type Story = StoryObj<typeof AvatarCore>;
+
+const defaultProps: TAvatarProps = {
+  name: "OD",
+};
+
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+const sizes = ["xxl", "xl", "l", "ml", "m", "s", "xs"] as any[];
```
haydencleary, line 16:
```suggestion
const sizes = ["xxl", "xl", "l", "ml", "m", "s", "xs"] as const;
```

Might help ?.

components/atoms/avatar/avatar.variants.ts
```diff
@@ -0,0 +1,110 @@
...
+      shape: "square",
+      class: {
+        base: "rounded",
+      },
+    },
+  ],
+  defaultVariants: {
+    size: "m",
+    shape: "round",
+    bg: "light",
```
haydencleary, line 108: Maybe use `container` instead of `bg` ?

components/atoms/tag/tag.types.ts
```diff
@@ -29,5 +29,5 @@ export interface TTagIconProps<C extends ElementType> extends TTagProps<C> {
 
 /** TODO refactor this with new avatar */
```
haydencleary, line 30: Can remove this now no ?

components/atoms/paper/paper.core.tsx
```diff
@@ -5,10 +5,20 @@ import { cn } from "src/utils/cn";
 import { TPaperProps } from "./paper.types";
 import { PaperCoreVariants } from "./paper.variants";
 
-export function PaperCore<C extends ElementType = "article">({ classNames, as, ...props }: TPaperProps<C>) {
+export function PaperCore<C extends ElementType = "article">({
+  classNames,
+  children,
+  as,
+  htmlProps,
```
haydencleary, line 12: Why do we need to expose `children` and `htmlProps` here ? `children` is a valid prop that can be spread onto a component.

haydencleary (changes requested):

alexbeno:
> Just for an avatar maybe NextUI isn't required ?

They do something cool on the fallback, if you passe name (like AB) and a broken src the name will fallback the src and if they don't have name the avatar have a default OD fallback

components/atoms/avatar/avatar.loading.tsx
alexbeno: Don't think for skeleton it's necessary but we can ?

components/atoms/avatar/avatar.variants.ts
alexbeno: Go for containerBackground ?

components/atoms/paper/paper.core.tsx
alexbeno: Let's talk together about html props

haydencleary (approved):

PR #2334, opened 2024-06-14
Add pagination

alexbeno:


components/molecules/pagination/pagination.core.tsx
```diff
@@ -0,0 +1,48 @@
...
+        size={"l"}
+        onClick={onPrev}
+        isDisabled={disablePrev}
+        hideText
+        startIcon={{ remixName: "ri-arrow-left-s-line" }}
+      />
+      <ButtonSecondaryLight size={"l"} as={"div"}>
+        {label}
+      </ButtonSecondaryLight>
+      <ButtonSecondaryLight
```
haydencleary, line 39: Shouldn't pagination accept a `variant` prop to define the button variant ? What happens if we want pagination with a primary or other variant ?

haydencleary: Same for the size

haydencleary: Should we allow the buttons to be in a loading state once clicked ?

haydencleary (changes requested):

components/molecules/pagination/pagination.core.tsx
alexbeno: Yes maybe we want a ButtonProps that will be spread inside they three button ? What do you think ?

haydencleary (approved):

PR #2335, opened 2024-06-16
Add input

alexbeno:


components/atoms/input/input.loading.tsx
```diff
@@ -0,0 +1,3 @@
+export function InputLoading() {
+  return <div />;
```
haydencleary, line 2: No skeleton ?

components/atoms/input/input.core.tsx
```diff
@@ -0,0 +1,50 @@
...
+      }}
+      label={label}
+      variant="bordered"
+      labelPlacement="outside-left"
+      startContent={startContent}
+      endContent={endContent}
+      isDisabled={disabled}
+      disabled={disabled}
+      onValueChange={handleChange}
+      onChange={onChange}
```
haydencleary, line 46: I don't think we need both, just the `onChange` should do.

components/atoms/input/input.core.tsx
```diff
@@ -0,0 +1,50 @@
...
+        helperWrapper: cn(slots.helperWrapper(), classNames?.helperWrapper),
+        description: cn(slots.description(), classNames?.description),
+      }}
+      label={label}
+      variant="bordered"
+      labelPlacement="outside-left"
+      startContent={startContent}
+      endContent={endContent}
+      isDisabled={disabled}
+      disabled={disabled}
```
haydencleary, line 44: Both necessary ?

components/atoms/input/input.core.tsx
```diff
@@ -0,0 +1,50 @@
...
+
+import { cn } from "src/utils/cn";
+
+import { TInputProps } from "./input.types";
+import { InputCoreVariants } from "./input.variants";
+
+export function InputCore({
+  classNames,
+  isError,
+  disabled,
```
haydencleary, line 11: I think it should be either `error` and `disabled` or `isError` and `isDisabled`. Slight preference for the prefixed version.

haydencleary (changes requested):

components/atoms/input/input.core.tsx
alexbeno: Next ui api.. :/

components/atoms/input/input.core.tsx
alexbeno: So isError and isDisabled ?

haydencleary (approved):

PR #2336, opened 2024-06-16
Add Badge

alexbeno:


components/atoms/badge/badge.core.tsx
```diff
@@ -0,0 +1,34 @@
...
+export function BadgeCore<C extends ElementType = "div">({
+  classNames,
+  as,
+  htmlProps,
+  children,
+  hideContent,
+  ...props
+}: TBadgeProps<C>) {
+  const Component = as || "div";
+  const { isFitContent, size, colors } = props;
```
haydencleary, line 20: Would use `fitContent` instead of `isFitContent`, as its not a state

components/atoms/badge/badge.variants.ts
```diff
@@ -0,0 +1,61 @@
...
+        base: "bg-brand1 text-text-1",
+      },
+      brand2: {
+        base: "bg-brand2 text-text-1",
+      },
+      brand3: {
+        base: "bg-brand3 text-text-1",
+      },
+      brand4: {
+        base: "bg-brand4 text-text-1",
```
haydencleary, line 24: As the brand color is applied to the background shouldn't the variant be `container` ? Although I'm not sure `outline` would fit here 🤷‍♂️

components/atoms/badge/badge.loading.tsx
```diff
@@ -0,0 +1,16 @@
...
+export function BadgeLoading<C extends ElementType = "div">({ size }: TBadgeProps<C>) {
+  const [w, h] = (() => {
+    if (size === "s") {
+      return [16, 16];
+    }
+
+    return [24, 24];
+  })();
+
+  return <SkeletonEl variant="rounded" width={w} height={h} className="rounded" />;
```
haydencleary, line 15: Don't need rounded variant here if you're overriding the class. I think we need to make a skeleton atom anyway

haydencleary (changes requested):

components/atoms/badge/badge.core.tsx
alexbeno: Good for me, like this logic if the props is stick to state it will be isProps else it should be props

components/atoms/badge/badge.variants.ts
alexbeno: We could anticipate and have a Color like this and a style ( fill and outline ) ?

haydencleary (approved):

PR #2337, opened 2024-06-16
Add Action link

alexbeno:


components/atoms/link/variants/link-default.tsx
```diff
@@ -0,0 +1,6 @@
+import { LinkCore } from "../link.core";
+import { TLinkProps } from "../link.types";
+
+export function Link({ ...props }: TLinkProps) {
+  return <LinkCore {...props} classNames={{}} />;
```
haydencleary, line 5:
```suggestion
  return <LinkCore {...props} />;
```

components/atoms/link/variants/link-button.tsx
```diff
@@ -0,0 +1,11 @@
+import { cn } from "src/utils/cn";
+
+import { LinkCoreVariants } from "components/atoms/link/link.variants";
+
+import { TLinkButtonProps } from "../link.types";
+
+export function LinkButton({ classNames, ...props }: TLinkButtonProps) {
+  const { color } = props;
```
haydencleary, line 8: Need to extract `htmlProps` otherwise `color` is still being passed to the button.

components/atoms/link/link.core.tsx
```diff
@@ -0,0 +1,25 @@
+import { cn } from "src/utils/cn";
+
+import { BaseLink } from "components/layout/base-link/base-link";
+import { Icon } from "components/layout/icon/icon";
+
+import { TLinkProps } from "./link.types";
+import { LinkCoreVariants } from "./link.variants";
+
+export function LinkCore({ classNames, children, ...props }: TLinkProps) {
+  const { color } = props;
```
haydencleary, line 10: Same as link-button, need to spread `htmlProps`

haydencleary (changes requested):

components/atoms/link/link.core.tsx
alexbeno: don't have html props here it extend the TbaseProps

components/atoms/link/variants/link-button.tsx
alexbeno: don't have html props here it extend the TbaseProps

haydencleary (approved):

PR #2346, opened 2024-06-19
Update remix icon and telegram icon

NeoxAzrot:
fix/remix-icons


haydencleary (commented):
Is the Telegram file still used ? Could delete it if not

haydencleary (approved):

PR #2398, opened 2024-06-25
B 1990 final

haydencleary:


api-client/resources/projects/fetch/get-project-issues.ts
```diff
@@ -0,0 +1,33 @@
...
+    fetchAdapter.setPathParams(pathParams);
+  }
+
+  if (queryParams) {
+    fetchAdapter.setParams(queryParams);
+  }
+
+  if (pathParams && queryParams) {
+    fetchAdapter.setTag(tags.get_project_issues({ pathParams, queryParams }));
+  }
```
alexbeno, line 30: This 3 conditions are weird ? if we don't have the path and queryParams we do the request but we don't tag it ?

alexbeno (approved):

PR #2414, opened 2024-06-27
refactor/molecules

haydencleary:
- **refactor(accordion): canOpenMultiple prop**
- **fix(accordion): updated prop usage**
- **refactor(pagination): adapters**
- **styles(pagination): hover and height**

### DIFF
--- a/components/molecules/accordion/accordion.stories.tsx
+++ b/components/molecules/accordion/accordion.stories.tsx
@@ -136,13 +136,13 @@ export const DefaultSelected: Story = {
 export const Multiple: Story = {
   parameters: {
     docs: {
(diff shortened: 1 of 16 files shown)


components/molecules/accordion/accordion.stories.tsx
```diff
@@ -136,13 +136,13 @@ export const DefaultSelected: Story = {
 export const Multiple: Story = {
   parameters: {
     docs: {
-      source: { code: "<Accordion selectionMode='multiple' />" },
+      source: { code: "<Accordion canOpenMultiple />" },
```
alexbeno, line 139: not fan of this naming isMultipleSelection or multiple would be more readable ?

alexbeno (approved):

PR #2417, opened 2024-06-27
Fix layout and fix graph re-render

alexbeno:


api-client/resources/applications/mutations/use-accept-application.ts
```diff
@@ -19,6 +24,7 @@ export function useAcceptApplication(
         queryKey: PROJECT_TAGS.good_first_issues(projectId),
         exact: false,
       });
+      router.push(NEXT_ROUTER.projects.details.root(projectSlug));
```
haydencleary, line 27: I don't think this should be done here. I would do the redirection where the mutation takes place.

app/p/[slug]/contributions/[contributionId]/features/contributor-details/components/total-earned/total-earned.tsx
```diff
@@ -33,9 +33,13 @@ export function TotalEarned({ githubId }: TTotalEarned.Props) {
       value: otherStats.reduce((acc, curr) => acc + curr.totalEarnedUsd, 0),
     };
 
-    return { data, otherData };
+    return [...data, otherData];
   }, [stats]);
 
+  const Graph = useMemo(() => {
```
Comment (haydencleary) [inline app/p/[slug]/contributions/[contributionId]/features/contributor-details/components/total-earned/total-earned.tsx:39]: Why is it necessary to use two `useMemo` hooks ?

app/p/[slug]/contributions/[contributionId]/page.tsx
```diff
@@ -26,12 +26,14 @@ function ContributionPage() {
...
     const applications = [...(newComers.applications || []), ...(projectMembers.applications || [])];
 
-    if (applications.length) {
+    if (applications.length && !selectedUser && !selectedApplication) {
       setSelectedUser(applications[0].applicant.githubUserId);
       setSelectedApplication(applications[0].id);
     }
   }, [newComers.applications, projectMembers.applications]);
 
+  console.log("REFECT 01");
```
Comment (haydencleary) [inline app/p/[slug]/contributions/[contributionId]/page.tsx]: Remove debug

haydencleary (changes requested):

app/p/[slug]/contributions/[contributionId]/features/contributor-details/components/total-earned/total-earned.tsx
Comment (alexbeno) [reply app/p/[slug]/contributions/[contributionId]/features/contributor-details/components/total-earned/total-earned.tsx:39]: When any data change the graph animation rerender due to nivo library so to handle that we specificy when the graph should be render, the first one only handle de data construction

haydencleary (approved):

PR #2454, opened 2024-07-04
Add categories in update project

alexbeno:


api-client/resources/projects/adapters.ts
```diff
@@ -10,6 +11,10 @@ const Adapters: { [key in Paths]: FetchAdapaterConstructor } = {
     url: "projects",
     method: "GET",
   },
+  get_all_categories: {
+    url: "project-categories",
```
haydencleary, line 15: Shouldn't this be inside a different resource ?

api-client/resources/projects/fetch/get-all-projects-categories.ts
```diff
@@ -0,0 +1,11 @@
+import { FetchAdapter } from "api-client/adapter/fetch/fetch-adapter";
+import { IFetchAdapater } from "api-client/adapter/fetch/fetch-adapter.types";
+import { ParametersInterface } from "api-client/types/parameters-interface";
+
+import adapters from "../adapters";
+import tags from "../tags";
+import { GetProjectCategories } from "../types";
+
+export function getAllProjectsCategories(_: ParametersInterface<object>): IFetchAdapater<GetProjectCategories> {
```
haydencleary, line 9: Why don't you remove the parameter as it's not used ?

api-client/resources/projects/adapters.ts
alexbeno: when we suggest this we have a route /project/categories same for create and update 

so i think for us the ressource is project

api-client/resources/projects/fetch/get-all-projects-categories.ts
alexbeno: to use the typeof inside the react query, if we remove this, we will have to refactor to extends the route

haydencleary (approved):

PR #2455, opened 2024-07-04
refacto permissions modals -> don't need to open drawer

pixelfact:


app/p/[slug]/features/good-first-issues/components/issue-card/components/apply-button/apply-button.tsx
```diff
@@ -1,15 +1,12 @@
...
-import { GrantPermission } from "components/features/grant-permission/grant-permission";
 import { Translate } from "components/layout/translate/translate";
 
 import { TApplyButton } from "./apply-button.types";
 
 export function ApplyButton({ hasApplied, onDrawerOpen }: TApplyButton.Props) {
-  const [isOpenGrantPermission, setIsOpenGrantPermission] = useState(false);
-
-  const renderButton = useMemo(() => {
+  return useMemo(() => {
```
Comment (haydencleary) [inline app/p/[slug]/features/good-first-issues/components/issue-card/components/apply-button/apply-button.tsx:9]: `useMemo` really required ? Just a simple condition and return null should do, no ?

haydencleary (approved):

PR #2516, opened 2024-07-19
revamp extend classes

alexbeno:


core/domain/issue/models/issue-model.ts
```diff
@@ -11,12 +11,12 @@ interface GithubIssueResponse extends _GithubIssueResponse {
...
 }
 
-export interface IssueInterface extends ListIssueInterface, GithubIssueResponse {}
-
-const ExtendedWithListIssue = extendMultipleClasses(mapApiToClass<GithubIssueResponse>(), ListIssue);
-
-export class Issue extends ExtendedWithListIssue implements IssueInterface {
-  constructor(protected readonly props: GithubIssueResponse) {
+export interface IssueInterface extends GithubIssueResponse {}
+export class Issue extends mapApiToClass<GithubIssueResponse>() implements IssueInterface {
```
haydencleary, line 15: `mapApiToClass` doesn't go inside `extendClasses` ?

core/domain/issue/models/issue-model.ts
```diff
@@ -11,12 +11,12 @@ interface GithubIssueResponse extends _GithubIssueResponse {
...
+  constructor(protected props: GithubIssueResponse) {
     super(props);
+    Object.assign(this, props);
```
haydencleary, line 18: This is still required here ?

core/domain/issue/models/issue-model.ts
alexbeno: If we want to get autocomplete of the props we should include it inside the extends

core/domain/issue/models/issue-model.ts
alexbeno: let's remove this

PR #2579, opened 2024-07-26
B 2239 integrate tunnel

alexbeno:


app/signup/components/account-already-exist/account-already-exist.tsx
```diff
@@ -0,0 +1,12 @@
+import { Typo } from "components/atoms/typo";
+import { Translate } from "components/layout/translate/translate";
+
+export function AccountAlreadyExist() {
+  return (
+    <Typo as="div" size="s" classNames={{ base: "py-2.5" }}>
+      <Translate token="v2.pages.signup.accountAlreadyExist.label" as={"span"} />
+      &nbsp;
+      <Translate token="v2.pages.signup.accountAlreadyExist.link" as={"a"} className="cursor-pointer text-text-2" />
```
haydencleary, line 9: Should be a `Link` or `BaseLink` to redirect to `/signup`

app/signup/onboarding/components/title/title.tsx
haydencleary, line 1: Could move this up a level as there are titles in the `/signup` page too, and refactor the UI there 👀

app/signup/onboarding/page.tsx
```diff
@@ -1,7 +1,95 @@
...
+      />
+    </div>
+  );
+}
+
 function OnboardingPage() {
-  return <SignupTemplate>ONBOARDING PAGE</SignupTemplate>;
+  const { data: user } = UserReactQueryAdapter.client.useGetMe({});
+
+  if (!user) return null;
```
haydencleary, line 38: No loading state ?

translations/v2/en/pages/signup/onboarding/onboarding-tunnel.json
```diff
@@ -0,0 +1,35 @@
+{
+  "title": "Complete Your Profile",
+  "content": "To personalize your experience, please follow these steps. The first step is mandatory, while the following ones are optional but highly recommended to enhance your experience.",
```
haydencleary, line 3: We say the first step is mandatory but really there are 2 mandatory steps.

core/domain/user/models/user-model.ts
```diff
@@ -23,7 +35,37 @@ export class User implements UserInterface {
...
+    return this.hasCompletedVerificationInformation;
+  }
+
+  isOnboardingWizardSeen() {
+    return this.hasSeenOnboardingWizard;
+  }
+
+  isTermsAndConditionsAccepted() {
+    return this.hasAcceptedLatestTermsAndConditions;
+  }
```
haydencleary, line 70: Why are you remapping these ?

app/signup/onboarding/page.tsx
alexbeno: we don't have it yet me could handle that we the funnel is done what do you think ?

core/domain/user/models/user-model.ts
alexbeno: We will use the steps multiple time in the app, to ensure that we don't do the same as we done on the isInstalled on organization. As this we can keep this logique inside the call and only refactore the model instead of the UI

translations/v2/en/pages/signup/onboarding/onboarding-tunnel.json
haydencleary: `The first two steps are mandatory, while the following ones are optional...`

core/domain/user/models/user-model.ts
haydencleary: Ok I see. I think we should still remove them all and add the methods as needed, otherwise we're going to find ourselves with unneeded code.

alexbeno: ok let's see in the future

haydencleary (approved):

PR #2704, opened 2024-10-23
Fix delete current user application

pixelfact:


haydencleary (changes requested):
Need to make a `getUserApplication` method in model

haydencleary (approved):

